PR #4, opened 2021-01-04
Add `draper` gem (& use to decorate user agent of Requests in admin)

davidrunger:


Gemfile.lock
```diff
@@ -25,9 +25,9 @@ GIT
 
 GIT
   remote: https://github.com/davidrunger/runger_style.git
-  revision: 6bbf9176b451637a905bd060fc1dad8b11b530dc
+  revision: a28a33be4827ab56a5da2851d95db15c7c7b95ab
```
davidrunger, line 28: this is required to avoid getting dinged by rubocop for `Rails/OrderById` (which we are violating in some specs added in this PR).

PR #5, opened 2021-01-04
Use `read_attribute` to access `method` of Request on admin show page

davidrunger:
fixes https://rollbar.com/davidjrunger/davidrunger/items/375/

### DIFF
--- a/app/decorators/request_decorator.rb
+++ b/app/decorators/request_decorator.rb
@@ -13,6 +13,10 @@ class RequestDecorator < ApplicationDecorator
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/decorators/request_decorator.rb
```diff
@@ -13,6 +13,10 @@ def pretty_user_agent
     end
   end
 
+  def to_s
+    "Request #{id} for #{handler}"
+  end
```
davidrunger, line 18: unrelated change, but what this does is to prettify the title of the show page

PR #32, opened 2021-01-06
Fix branch name in note added to log entry for build time

davidrunger:


davidrunger:
I don't think this is possible, per se. I am going to open a new PR with a better branch name, etc.

PR #42, opened 2021-01-07
Begin quiz game implementation (create Quiz, QuizParticipation)

davidrunger:
The interface currently isn't "realtime" at all; making it so (using [Hotwire][1]) will be our next step.

[1]: https://hotwire.dev/


app/assets/stylesheets/styles.scss
```diff
@@ -33,3 +33,7 @@ html {
 textarea {
   font-family: sans-serif;
 }
+
+a {
+  text-decoration: none;
+}
```
davidrunger, line 39: (This is required because of the switch from reset.css to normalize.css in this PR.)

PR #44, opened 2021-01-08
Require opting in to Turbo JavaScript (i.e. disable Turbo by default)

davidrunger:
Turbo was breaking lots of stuff in the parts of the application that weren't built with Turbo awareness in mind. (For example, links from the home page to the groceries app were doing nothing.)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,6 +8,7 @@ class ApplicationController < ActionController::Base
   include Redirectable
   include RequestRecordable
   include TokenAuthenticatable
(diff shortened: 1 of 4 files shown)


davidrunger:
Instead of requiring opt-in, we could instead require opt-out via  `<meta name="turbo-visit-control" content="reload">`. https://turbo.hotwire.dev/handbook/drive#ensuring-specific-pages-trigger-a-full-reload

However, since currently more pages want Turbo disabled than enabled, I think that the opt-in approach is good. We can revisit this in the future, if more controllers/pages start to use Turbo.

PR #78, opened 2021-01-16
Implement quiz gameplay (answer question, show answers, advance, etc)

davidrunger:
This is a _very_ rough initial implementation. We will be doing further iteration!


app/controllers/api/log_entries_controller.rb
```diff
@@ -6,7 +6,7 @@ def create
     log = (current_user || auth_token_user).logs.find(params.dig(:log_entry, :log_id))
     @log_entry = log.log_entries.build(log_entry_params)
     if @log_entry.valid?
-      LogEntries::Save.new(log_entry: @log_entry).execute
+      LogEntries::Save.new(log_entry: @log_entry).run!
```
davidrunger, line 9: (Unrelated change; we should always be calling `#run` or `#run!` (provided by `active_actions`), rather than our own implemented `#execute` method.)

app/views/layouts/application.haml
```diff
@@ -26,9 +26,6 @@
       window.davidrunger = {env: '#{Rails.env}'};
       window.davidrunger.bootstrap = JSON.parse("#{raw(escape_javascript((@bootstrap_data || {}).to_json))}");
 
-    - if @turbo_enabled
-      = stimulus_include_tags
-      = turbo_include_tags
```
davidrunger, line 31: Instead, we're now including `@hotwired/turbo` into a `quizzes.js` pack and loading that in `quiz#show`.

app/views/layouts/application.haml
```diff
@@ -26,9 +26,6 @@
       window.davidrunger = {env: '#{Rails.env}'};
       window.davidrunger.bootstrap = JSON.parse("#{raw(escape_javascript((@bootstrap_data || {}).to_json))}");
 
-    - if @turbo_enabled
-      = stimulus_include_tags
```
davidrunger, line 30: I've decided that it doesn't seem _too_ likely that I'll be using stimulus anytime soon (though I still might...), so I'm removing all of the stimulus references for now.

PR #106, opened 2021-01-20
Allow backdating log entries when creating them

davidrunger:
This can be useful in order to enter a log entry after a delay has passed since the actual measurement was taken / event occurred.

### DIFF
--- a/app/controllers/api/log_entries_controller.rb
+++ b/app/controllers/api/log_entries_controller.rb
@@ -73,6 +73,6 @@ class Api::LogEntriesController < ApplicationController
   private
 
   def log_entry_params
-    params.require(:log_entry).permit(:data, :note)
+    params.require(:log_entry).permit(:created_at, :data, :note)
   end
 end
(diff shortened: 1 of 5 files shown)


app/controllers/api/log_entries_controller.rb
```diff
@@ -73,6 +73,6 @@ def index
   private
 
   def log_entry_params
-    params.require(:log_entry).permit(:data, :note)
+    params.require(:log_entry).permit(:created_at, :data, :note)
```
davidrunger, line 76: Ideally, we'd create a new column (`timestamp` or something) rather than overriding the `created_at` timestamp in this inaccurate/misleading way, but, for the sake of simplicity, I'll do it this way.

app/javascript/logs/components/data_renderers/text_log.vue
```diff
@@ -113,7 +113,7 @@ ol {
   margin-left: 5px;
 }
 
-table {
+table.text-log-table {
```
davidrunger, line 116: (These styling rules were being applied to the table that is used by the Element UI date-time picker, causing a visual bug. By adding a specific class, these styling rules no longer apply to the table that is used by the Element UI date-time picker, resolving that bug.)

PR #109, opened 2021-01-22
Set `config.active_record.strict_loading_by_default` to `true`

davidrunger:
See https://github.com/rails/rails/pull/ 39491 .


davidrunger:
This setting is way too strict. I'm going with https://github.com/davidrunger/david_runger/pull/110 instead.

PR #110, opened 2021-01-22
Add `bullet` gem to guard against N+1 queries and over-eager-loading

davidrunger:
`bullet` will only be active in `development` and `test` (not `production`).


app/controllers/logs_controller.rb
```diff
@@ -27,7 +27,7 @@ def index
     bootstrap(
       current_user: UserSerializer.new(current_user),
       logs: ActiveModel::Serializer::CollectionSerializer.new(
-        logs.includes(:log_shares),
```
davidrunger, line 30: We were doing unnecessary eager loading here in the case that a user was viewing a log that had been shared with him.

PR #119, opened 2021-01-24
Add quizzes #index page and instructions to #new page

davidrunger:
Also, set up `bullet` in tests correctly and fix some N+1 queries.

Also, refactor how we set <body> class(es) / container class(es).


app/views/quiz_questions/_closed.html.haml
```diff
@@ -2,7 +2,7 @@
 - question.answers.includes(:answering_participations).decorate.each_with_index do |answer, index|
   .pl1
     %span{class: answer.is_correct? ? 'bold' : ''} #{(index + 65).chr}. #{answer.content}
-    - if answer.answering_participations.any?
+    - if answer.answering_participations.present?
```
davidrunger, line 5: `any?` was triggering an N+1 query (even though we already preloaded the `answering_participations`); `present?` doesn't trigger a query.

config/environments/test.rb
```diff
@@ -40,6 +41,9 @@
   # Print deprecation notices to the stderr.
   config.active_support.deprecation = :stderr
 
+  # Highlight code that triggered database queries in logs.
+  config.active_record.verbose_query_logs = true
```
davidrunger, line 45: this makes it a little easier to find the source of `bullet` errors in specs

PR #121, opened 2021-01-25
Move log entry storage to "delegated types"

davidrunger:


davidrunger:
This destroys the performance of e.g. `Api::LogEntriesController#index` and should probably not be merged. With this change, we're looking at more like ~6s response times locally for one of my logs with a lot of log entries (versus of ~600ms response times for the same log locally on `master`).

PR #131, opened 2021-01-28
Use Postgres (rather than AMS) to render log entries JSON

davidrunger:
(AMS = ActiveModel::Serializers)

This is much faster! When using production-ish server settings on my local machine, this change takes the response time for `/api/log_entries` from ~500ms to ~50ms. Nice improvement!

This article was a helpful guide: https://bigbinary.com/blog/generating-json-using-postgresql-json-function .

### DIFF
--- a/app/controllers/api/log_entries_controller.rb
+++ b/app/controllers/api/log_entries_controller.rb
@@ -50,29 +50,64 @@ class Api::LogEntriesController < ApplicationController
     authorize(LogEntry)
     log_id = params['log_id']
 
-    log_entries =
+    log_entry_json_strings =
       if log_id.present?
         log = Log.find(log_id)
         authorize(log, :show?)
-        log.log_entries
+        log_entry_json_strings_for_log(log)
       else
-        logs =
-          current_user.logs.
-            includes(:number_log_entries, :text_log_entries).
-            to_a
-        logs.map!(&:log_entries).flatten!
-        logs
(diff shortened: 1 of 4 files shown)


davidrunger:
The first thing that I tried (just on my local machine) was actually to build a [Kemal](https://kemalcr.com/) server (in the [Crystal](https://crystal-lang.org/) programming language), which I then reverse-proxied to (using [this fork](https://github.com/davidrunger/rails-reverse-proxy) of [`rails-reverse-proxy`](https://github.com/davidrunger/rails-reverse-proxy)) from `Api::LogEntriesController#index` (after performing authentication/authorization in Rails). That solution also gave a very large performance benefit (comparable or maybe even better than the performance benefit from the change in this PR), but that approach would have been be a lot more complicated to test, deploy, maintain, etc.

davidrunger:
GitHub actions is having an outage right now, but I ran the specs on my local machine, and they all passed. Merging.

davidrunger:
Here are some more-or-less representative times from production. (Locally, the times are lower; I think that probably mostly has to do with network time in production that doesn't happen locally.)

## Before

image

## After

image

davidrunger:
NewRelic makes the performance gains more clear:

image

davidrunger:
Here's a graph of `db` + `view` time as reported by Rails and stored in the most recent 68 `Request`s for all of my logs that I've made:

image

PR #132, opened 2021-01-28
Add `--no-pager` option when running `brakeman`

davidrunger:


davidrunger:
GitHub actions is having an outage right now, but I ran the specs on my local machine, and they all passed. Merging.

PR #165, opened 2021-01-30
Automagically import only required functions via babel-plugin-lodash

davidrunger:
I verified using `webpack-bundle-analyzer` that this is indeed working as expected/intended. The `lodash` import is broken up into smaller modules and only imports those modules which are actually used.

image

### DIFF
--- a/app/javascript/groceries/store.js
+++ b/app/javascript/groceries/store.js
@@ -91,7 +91,7 @@ const getters = {
     if (!state.stores) return null;
 
     return (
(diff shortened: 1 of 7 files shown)


app/javascript/groceries/store.js
```diff
@@ -91,7 +91,7 @@ const getters = {
     if (!state.stores) return null;
 
     return (
-      _(state.stores).filter('viewed_at').sortBy(['viewed_at']).last() ||
```
davidrunger, line 94: this "chaining" is not supported by `babel-plugin-lodash`.

PR #169, opened 2021-01-30
Give the homepage a monospace font

davidrunger:
It gives the page actually a somewhat interesting look and saves us from loading an external stylesheet from Google and then loading the webfont required by that stylesheet.


davidrunger:
This finally got me to a glorious 100-100-100 score in Chrome's Lighthouse tool:

image

**Booyah!**

PR #171, opened 2021-01-31
Use cuprite rather than selenium to drive feature tests

davidrunger:


davidrunger:
This actually ended up being a tiny bit slower than Selenium when tested on my local machine (like 17.5 seconds for Cuprite vs 16.0 seconds for Selenium), and it required two changes to fix flakiness (see the spec files changed in this PR), so I'm not going to merge this. I had thought (as claimed here https://evilmartians.com/chronicles/system-of-a-test-setting-up-end-to-end-rails-testing ) that Cuprite would be faster and that might make it worthwhile, but the reality is that (somewhat surprisingly) it's not.

PR #177, opened 2021-02-01
Store `BannedPathFragment`s in the database (i.e. no longer in code)

davidrunger:
This will make it much easier to add new banned path fragments.

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,23 +1,7 @@
 # frozen_string_literal: true
 
 class Rack::Attack
-  BANNED_PATH_FRAGMENTS = Set.new(%w[
-    administrator
-    env
-    git
-    old-wp
-    passwd
-    php
-    phpformbuilder
-    phpmyadmin
-    phpmyadmin
-    phpunit
-    plugins
-    wordpress
-    wp
-    wp1
-    wp2
-  ].map(&:freeze)).freeze
+  PATH_FRAGMENT_SEPARATOR_REGEX = %r{/|\.|\?|-|_}.freeze
   PENTESTERS_PREFIX = 'pentesters-'
   PENTESTING_FINDTIME = 1.day.freeze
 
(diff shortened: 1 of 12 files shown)


spec/models/ip_block_spec.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-RSpec.describe IpBlock, type: :model do
```
davidrunger, line 3: unrelated change; i noticed that this unnecessary type annotation is here (unnecessary because we have `config.infer_spec_type_from_file_location!` in `spec/spec_helper.rb`).

config/initializers/rack_attack.rb
```diff
@@ -1,23 +1,7 @@
...
-    phpformbuilder
-    phpmyadmin
-    phpmyadmin
-    phpunit
-    plugins
-    wordpress
-    wp
-    wp1
-    wp2
-  ].map(&:freeze)).freeze
```
davidrunger, line 20: removing from the codebase and ceasing to maintain / add to this list of banned path fragments (instead maintaining said list via the admin web interface) is the motivation for this change.

db/datamigrate/populate_banned_path_fragments.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+def populate_banned_path_fragments
+  %w[
+    administrator
+    env
+    git
+    passwd
```
davidrunger, line 8: note that I've removed `old-wp` from this list, since that's not a valid fragment value (since it has `-` in it, which means a path containing that would be split into `old` and `wp` anyway (and we have a `wp` banned fragment))

spec/features/home_spec.rb
```diff
@@ -63,6 +63,7 @@
       Full stack web developer
     HEADLINE
 
+    sleep(2) # allow some extra time for CSS `background-image`s to load and be rendered
```
davidrunger, line 66: unrelated change (I'm pretty sure?), but for whatever reason Percy seemed determine to flake (showing the home page "parallax" background images in the `master` reference screenshots, but not in the screenshots generated from this PR)

davidrunger: i initially tried `0.5` seconds, but it still flaked even then; hopefully `2` seconds will be enough to prevent flakiness.

PR #179, opened 2021-02-01
Validate uniqueness of `BannedPathFragment#value`

davidrunger:


spec/models/banned_path_fragment_spec.rb
```diff
@@ -4,7 +4,7 @@
   subject(:banned_path_fragment) { BannedPathFragment.new }
 
   it { is_expected.to validate_presence_of(:value) }
-  it { is_expected.to allow_value('wordpress').for(:value) }
-  it { is_expected.not_to allow_value('/wordpress').for(:value) }
+  it { is_expected.to allow_value('phpmyadmin').for(:value) }
+  it { is_expected.not_to allow_value('/phpmyadmin').for(:value) }
```
davidrunger, line 8: I needed to change these from `wordpress` to `phpmyadmin` because there is already a `wordpress` one created in `spec/support/fixture_builder.rb`, which was making the `wordpress` one that we were instantiating here invalid.

PR #200, opened 2021-02-02
Use `moment-locales-webpack-plugin` to shrink size of logs JS bundle

davidrunger:
This shrinks the production `logs_app` bundle from 1.1M (237K w/ brotli compression) to 785K (198K w/ brotli).

# Before

image

# After

image

### DIFF
--- a/config/webpack/shared.js
+++ b/config/webpack/shared.js
@@ -2,8 +2,9 @@ const webpack = require('webpack');
 const { basename, dirname, join, relative, resolve } = require('path');
 const { sync } = require('glob');
 const extname = require('path-complete-extname');
-const { settings } = require('./configuration.js');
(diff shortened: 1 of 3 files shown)


config/webpack/shared.js
```diff
@@ -2,8 +2,9 @@ const webpack = require('webpack');
 const { basename, dirname, join, relative, resolve } = require('path');
 const { sync } = require('glob');
 const extname = require('path-complete-extname');
-const { settings } = require('./configuration.js');
 const { VueLoaderPlugin } = require('vue-loader');
+const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
+const { settings } = require('./configuration.js');
```
davidrunger, line 7: eslint asked that this relative import be moved down

PR #238, opened 2021-02-07
Add and use `secure_headers` gem with default settings

davidrunger:


bin/release-tasks
```diff
@@ -5,23 +5,23 @@ migrate() {
...
   done
   [ $NEXT_WAIT_TIME -lt 10 ]
+
+  echo "Purging prerendered pages..."
+  bin/rails runner bin/purge-prerendered-pages
+  echo "... done."
+
+  echo "Enqueueing job to track asset sizes..."
+  bin/rails runner "TrackAssetSizes.perform_in(2.minutes)"
+  echo "... done."
```
davidrunger: This is mostly an unrelated change, but (since I'm testing this by deploying a review app), I realized that we don't want/need these tasks to run when deploying a review app. I'm unsure why I didn't think of this when I initially added the first of these tasks when setting up prerendering.

app/views/layouts/application.haml
```diff
@@ -22,7 +22,7 @@
       %meta{property: 'og:description', content: @description}
 
     = csrf_meta_tags
-    %script{type: "text/javascript"}
```
davidrunger, line 25: mostly unrelated change, except that i realized while working on this change that `type: "text/javascript"}` is not needed

config/initializers/secure_headers.rb
```diff
@@ -0,0 +1,20 @@
...
+SecureHeaders::Configuration.default do |config|
+  # rubocop:disable Lint/PercentStringArray
+  config.csp = {
+    base_uri: %w['self'],
+    default_src: %w['self'],
+    connect_src: %w['self' wss:],
+    font_src: %w['self' cdn.rawgit.com],
+    img_src: %w['self' data:],
+    object_src: %w['none'],
+    script_src: %w['self' 'unsafe-inline' 'unsafe-eval'],
```
davidrunger: I tried to get rid of the `'unsafe-inline'` by using a nonce instead, but unfortunately it seems that we cannot add a nonce (that is respected) to cover the `onload` handler for our font preloading `<link>`. By using `'unsafe-inline'`, I'm favoring performance over security.

app/views/layouts/application.haml
davidrunger: nevermind; this awesome answer explained how to do it: https://stackoverflow.com/a/63984111/4009384

davidrunger: actually, that answer doesn't quite work, i think, because the font can load before the script gets parsed, in which case the `load` event handler doesn't get called, and so the font doesn't get registered / converted to a stylesheet. i fixed it by manually adding the `href` attribute after registering the load listener. a lot of work is required to squeeze out that last 1% of security and performance together...

PR #239, opened 2021-02-07
Revert "Add and use `secure_headers` gem with default settings"

davidrunger:
Reverts davidrunger/david_runger#238


davidrunger:
I'm reverting this because:
1. I have realized that it doesn't work with prerendering (since the nonces don't match); this could maybe be fixed, but I don't think it's worth the effort/complexity
2. this breaks the flipper web UI and causes at least a warning in the sidekiq web UI; I could try to serve a different CSP for those pages, but it's not immediately obvious how to do so, since those are mounted rack apps rather than controller actions

PR #311, opened 2021-02-19
Update Redis connection counting logic and allocations

davidrunger:


davidrunger:
CodeCov is having issues. It claims this change reduces coverage, but I believe that's a lie. Merging.

PR #619, opened 2021-05-19
Revert "Add git registry for dependabot to access private git repos"

davidrunger:
Reverts davidrunger/david_runger#615


davidrunger:
Unfortunately, this functionality doesn't seem to be supported for `bundler`. See https://github.community/t/dependabot-v2-bundler-ruby-is-there-a-plan-to-support-private-repositories/177577 .

PR #731, opened 2021-07-19
Update gems

davidrunger:
`bundle update`


davidrunger:
I think we need to wait for this issue to be fixed: https://github.com/percy/cli/issues/ 429

PR #901, opened 2021-10-26
Update gems

davidrunger:
`bundle update`


davidrunger:
Skipping this because of `Layout/DotPosition` RuboCop issues; hopefully most/all of these are a bug that will be fixed in a future release.

PR #923, opened 2021-11-09
Update gems

davidrunger:
`bundle update`

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "stylelint-scss": "^3.21.0",
     "stylelint-webpack-plugin": "^2.2.2",
     "vite": "^2.6.13",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -96,8 +96,8 @@
     "stylelint-config-standard": "^22.0.0",
     "stylelint-scss": "^3.21.0",
     "stylelint-webpack-plugin": "^2.2.2",
-    "vite": "^2.6.11",
-    "vite-plugin-ruby": "^3.0.1",
+    "vite": "^2.6.13",
+    "vite-plugin-ruby": "^3.0.2",
```
davidrunger: These changes came from `bundle exec vite upgrade`, which was recommended in the post-install messages of `bundle update`.

PR #1239, opened 2022-05-12
Add policy scope for logs

davidrunger:
This isn't actually used in the application code at this time, but I think it's a good idea to be consistent having a `Scope` for all policies and it makes the `ApplicationPolicy` spec read better.


davidrunger:
Going with https://github.com/davidrunger/david_runger/pull/1240 instead.

PR #1297, opened 2022-05-25
Use vite for production asset compilation and drop webpack(er)

davidrunger:


app/javascript/packs/quizzes.js
```diff
@@ -4,7 +4,6 @@ import { Turbo } from '@hotwired/turbo-rails';
 import actionCableConsumer from '@/channels/consumer';
 import { loadAsyncPartials } from '@/lib/async_partial';
 
-Rails.start();
```
davidrunger, line 7: We were getting a warning now that we are compiling with vite. `Rails.start()` was being called twice (once here and once automatically). I'm not sure why it wasn't being called automatically before, but it is now.

spec/spec_helper.rb
```diff
@@ -37,7 +37,7 @@
 Sidekiq.logger = Rails.logger
 
 WebMock.enable!
-WebMock.disable_net_connect!(allow_localhost: true)
+WebMock.disable_net_connect!(allow_localhost: true, net_http_connect_on_start: true)
```
davidrunger, line 40: https://stackoverflow.com/a/65946077/4009384

PR #1319, opened 2022-05-29
Store asset size timeseries in Postgres rather than Redis

davidrunger:
Storing them in Redis was using too much memory when reading the values
because the timestamps and values were stored as strings. By storing
them in a Postgres JSON array, we can leave them as integers and floats,
saving memory when reading the values. I think inserts will be a bit
slower and use more memory, but that's okay.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -18,7 +18,7 @@ GIT
 
 GIT
   remote: https://github.com/davidrunger/runger_style.git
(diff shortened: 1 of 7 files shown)


davidrunger:
Changing branch name. Closing.

davidrunger:
See https://github.com/davidrunger/david_runger/pull/1320.

PR #1326, opened 2022-06-04
Store secrets in Rails credentials file

davidrunger:


.github/workflows/ruby.yml
```diff
@@ -59,6 +59,7 @@ jobs:
       run: bin/pallets --verbose -r ./bin/run-tests.rb
       env:
         CI: true
+        AWS_EC2_METADATA_DISABLED: true
```
davidrunger, line 62: Without this, specs run differently locally vs on GitHub Actions CI when AWS credentials are not present. I believe that this difference is attributable to the fact that GitHub Actions probably runs on AWS machines.

PR #1393, opened 2022-06-14
Add `sleep`s in spec before and after signing in quiz participant

davidrunger:
Hopefully this will reduce flakiness. A screenshot showed the login screen one time when this spec flaked.


davidrunger:
Closing this for a better branch name and a better approach. See https://github.com/davidrunger/david_runger/pull/1394.

PR #1403, opened 2022-06-15
Remove `sass` JS package

davidrunger:
It doesn't seem that we need this anymore.


davidrunger:
Actually, we do need it.

PR #1433, opened 2022-06-23
Switch redis.tls_url credential from Heroku to Redis Labs

davidrunger:
Also, delete redis.url credential, as we aren't using it.


davidrunger:
I can only get this to work with `redis`, not `rediss` (TLS-enabled). Since that doesn't seem secure, I'm going to close this.

PR #1439, opened 2022-06-25
Use link API for el-button (not deprecated `type='text'`)

davidrunger:


app/controllers/workouts_controller.rb
```diff
@@ -9,7 +9,7 @@ def index
       current_user: UserSerializer.new(current_user),
       workouts:
         ActiveModel::Serializer::CollectionSerializer.new(
-          current_user.workouts.
+          current_user.workouts.includes(:user).
```
davidrunger, line 12: Testing the workouts app triggered bullet to require eager-loading here.

PR #1692, opened 2022-09-19
Use redis-client for Sidekiq Redis client

davidrunger:
We are now installing `sidekiq-scheduler` from GitHub to get these fixes (which are not yet released) https://github.com/sidekiq-scheduler/sidekiq-scheduler/pull/ 404 .


davidrunger:
We still are also using the `redis` gem because our Flipper adapter and `redlock` still depend on it (and it's still listed as a dependency of Sidekiq, so we'd still be installing the gem, anyway).

PR #1696, opened 2022-09-20
Only require `activeadmin` in web processes

davidrunger:
This should save some memory elsewhere (e.g. in Sidekiq processes).


davidrunger:
I think that this is a bad idea. Closing.

PR #1723, opened 2022-09-24
Run `bin/vite build` after `assets:precompile` rake task

davidrunger:
Having just set `VITE_RUBY_SKIP_ASSETS_PRECOMPILE_EXTENSION=true` on dokku, we now need to ourselves enhance the `assets:precompile` rake task to build our JavaScript bundles (now including the main app bundles in addition to the admin bundles).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,11 +6,3 @@
 require_relative 'config/application'
 
 Rails.application.load_tasks
-
-if Rails.env.production?
-  # https://github.com/heroku/heroku-buildpack-ruby/pull/892#issuecomment-548897899
-  assets_precompile_task = Rake.application.tasks.find { |task| task.name == 'assets:precompile' }
-  if assets_precompile_task.present?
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -6,11 +6,3 @@
...
 
 Rails.application.load_tasks
-
-if Rails.env.production?
-  # https://github.com/heroku/heroku-buildpack-ruby/pull/892#issuecomment-548897899
-  assets_precompile_task = Rake.application.tasks.find { |task| task.name == 'assets:precompile' }
-  if assets_precompile_task.present?
-    assets_precompile_task.prerequisites.delete('yarn:install')
-  end
-end
```
davidrunger, line 16: The code being deleted here wasn't doing anything anymore. `yarn:install` was not anymore a prerequisite of `assets:precompile`. Rather, `yarn install` was being run because `vite:install_dependencies` is added to the `assets:precompile` rake task (which, as a side note, I think is now provided by `sprockets` rather than by `rails` itself).

PR #1761, opened 2022-09-27
Move testing `require`s and other code into per-test-type blocks

davidrunger:


davidrunger:
Somewhat surprisingly, this actually seems to slow down tests. Closing.

PR #1769, opened 2022-09-29
Convert release-tasks script from bash to using Rails runner

davidrunger:
This should be faster than separately invoking various Rails commands.


davidrunger:
Woops, my automated script merged this unintentionally before the GitHub checks had passed due to an oversight in the script's implementation.

davidrunger:
Fortunately, the tests did pass, though.

PR #1828, opened 2022-10-04
Block external requests by default in feature tests

davidrunger:


davidrunger:
Unfortunately, this seems to slow down the feature spec execution time from ~38 seconds to ~48 seconds, which I don't think is worth it. Closing for now.

PR #1836, opened 2022-10-05
Source pallets gem from GitHub fork

davidrunger:
This removes deprecation warnings that are printed when running tests. (See https://github.com/davidrunger/pallets/pull/1 .)

There is already a PR for this at the main pallets repo: https://github.com/linkyndy/pallets/pull/ 64

Unfortunately, I guess it looks like maybe `pallets` is no longer being maintained, though.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -93,7 +93,7 @@ group :test do
   gem 'json-schema'
   gem 'launchy'
   gem 'nokogiri'
-  gem 'pallets'
+  gem 'pallets', github: 'davidrunger/pallets'
   gem 'percy-capybara'
   gem 'rails-controller-testing'
   gem 'rspec-instafail', require: false


davidrunger:
This is the deprecation warning that we are addressing:

```
Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.

redis.multi do
  redis.get("key")
end

should be replaced by

redis.multi do |pipeline|
  pipeline.get("key")
end

(called from /home/runner/work/david_runger/david_runger/vendor/bundle/ruby/3.1.0/gems/pallets-0.11.0/lib/pallets/backends/redis.rb:99:in `block in run_workflow'}
```

davidrunger:
GitHub actions is down. I'm going to merge this without checks.

PR #1837, opened 2022-10-05
Also check ENV vars in storage.yml for AWS access key id and secret

davidrunger:
This will make it easier to use the `:amazon` storage service locally.


davidrunger:
GitHub actions is having problems. Ran checks locally and it passes. Merging.

PR #1938, opened 2022-10-16
Include Rollbar username davidjrunger in deploys post body

davidrunger:
https://rollbar.com/davidjrunger/davidrunger/items/519/


davidrunger:
I'm hoping that this will make Rollbar auto-resolving work. It currently doesn't seem to be working.

PR #2000, opened 2022-10-22
Add CspReportDecorator#to_s to improve the title on admin show page

davidrunger:


.irbrc
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-home_irbrc_path = "#{ENV['HOME']}/.irbrc.rb"
+home_irbrc_path = "#{Dir.home}/.irbrc.rb"
```
davidrunger, line 3: (This fixes a rubocop error that was missed previously because we weren't considering `.irbrc` as a Ruby file, so Rubocop wasn't executed when `ENV['HOME']` was added.)

PR #2009, opened 2022-10-22
Create `useSubscription` composable to unsubscribe automatically

davidrunger:


app/javascript/lib/event_bus.js
```diff
@@ -5,7 +5,9 @@
 // returns a function that unsubscribes from the event
 export function on(eventName, callback) {
   window.addEventListener(eventName, callback);
-  return () => { window.removeEventListener(eventName, callback); };
+  return function unsubscribe() {
```
davidrunger, line 8: (This is just so that the function is named `unsubscribe` rather than `on`, which clarifies debugging output in the devtools console.)

PR #2112, opened 2022-11-03
Replace `http` with `https` wherever possible

davidrunger:


davidrunger:
CodeCov seems to be going slowly. Coverage is still 100%, I think. Merging.

PR #2120, opened 2022-11-03
Refactor based on RubyCritic critique

davidrunger:
Mostly this addresses its complaints about calling a method multiple times within a method.

### DIFF
--- a/lib/redis_options.rb
+++ b/lib/redis_options.rb
@@ -16,8 +16,8 @@ class RedisOptions
 
   memoize \
   def url_without_db
-    case ENV.fetch('RAILS_ENV', nil)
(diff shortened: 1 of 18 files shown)


lib/redis_options.rb
```diff
@@ -16,8 +16,8 @@ def url
 
   memoize \
   def url_without_db
-    case ENV.fetch('RAILS_ENV', nil)
-    when 'development', 'test', nil then 'redis://localhost:6379'
```
davidrunger, line 20: I think that all of the other changes in this PR are pure refactors (code changes not expected to change behavior). I think that this is the one change that actually changes behavior (or, at least, would change behavior, if RAILS_ENV is ever `nil`, but I think it never should be, which is why I'm making this change).

PR #2216, opened 2022-11-13
Convert all `forEach` iterations to `for-of` loops

davidrunger:
I think that the `for-of` syntax is a little cleaner and easier to read.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -46,6 +46,12 @@ module.exports = {
         message: 'Use lodash-es.',
       }],
     }],
+    'no-restricted-syntax': [
+      'error',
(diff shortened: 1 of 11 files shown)


.eslintrc.js
```diff
@@ -46,6 +46,12 @@ module.exports = {
         message: 'Use lodash-es.',
       }],
     }],
+    'no-restricted-syntax': [
+      'error',
+      'ForInStatement',
+      'LabeledStatement',
+      'WithStatement',
+    ],
```
davidrunger, line 54: airbnb bans `ForOfStatement`; this overrides that and allows it.

PR #2414, opened 2023-01-17
Don't duplicate stores in groceries store when resuming WS connection

davidrunger:
This fixes a bug wherein, after the WebSocket connection was lost and then reestablished, stores (and their items) could/would be duplicated within the check-in interface.

### DIFF
--- a/app/javascript/groceries/store.js
+++ b/app/javascript/groceries/store.js
@@ -96,19 +96,22 @@ const actions = {
   },
 
   pullStoreData() {
-    kyApi.get(Routes.api_stores_path()).json().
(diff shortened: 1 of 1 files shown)


davidrunger:
Unfortunately, it's hard to write a test for this, because doing so would involve disconnecting and then reconnecting the WebSocket connection in a test, which I tried to do using various Chrome/Selenium APIs (`emulateNetworkConditions`, etc.), but I couldn't get that working, and I'm not sure that it's possible to get working right now; there are some bug reports for Chromium that make me think that this might be a bug / missing feature in Chromium.

PR #2756, opened 2023-05-08
Update gems

davidrunger:
`bundle update`


davidrunger:
Note that we are holding `selenium-webdriver` back at 4.9.0 because 4.9.1 has a change that is incompatible (https://github.com/teamcapybara/capybara/issues/ 2666) with Capybara.

PR #2832, opened 2023-05-29
Include `--frozen-lockfile` option for CI pnpm install command

davidrunger:
This is recommended by these docs: https://github.com/actions/setup-node/blob/869f4dd0c7f320ae834c2724d92a364de3893c24/docs/advanced-usage.md

> Ensure that `pnpm-lock.yaml` is always committed, when on CI pass `--frozen-lockfile` to `pnpm install` when installing packages.


davidrunger:
According to [these docs](https://pnpm.io/cli/install#--frozen-lockfile), this was probably already happening by default, since we have a `CI` environment variable present in CI.

PR #2850, opened 2023-06-05
Update JavaScript packages

davidrunger:
`pnpm up`


davidrunger:
We are locking `cosmiconfig` to 8.1.3 because of a bug.

See:
- https://github.com/stylelint/stylelint/issues/ 6898
- https://github.com/cosmiconfig/cosmiconfig/issues/ 307

PR #3591, opened 2023-12-04
Fix pnpm lockfile

davidrunger:
`pnpm install`

I guess this got broken by a recent dependabot PR (maybe because of a badly handled merge conflict (or something that wasn't actually surfaced as a merge conflict but which disagreed between two PRs)).


davidrunger:
Yeah, I think there might have been a bad merge between #3585 and #3586 (which e.g. both mention `nanoid` in the diff).

PR #4402, opened 2024-06-17
[groceries] Fix Tailwind styling of check-in modal

davidrunger:
... plus global find-and-replaces.


app/javascript/groceries/components/store.vue
```diff
@@ -1,6 +1,6 @@
 <template lang="pug">
 .store-container.overflow-auto.hidden-scrollbars.pt-2.pb-4.pl-8.pr-4
-  h2.store-name.bold.my-4
```
davidrunger, line 3: `bold` is not needed here (anymore?) since we have applied bolding to all `h2`s.

PR #4427, opened 2024-06-21
Add "packageManager": "pnpm@9.4.0" to package.json

davidrunger:
Hopefully this will fix dependabot JavaScript updates. https://github.com/dependabot/dependabot-core/issues/ 9684#issuecomment-2171818307


davidrunger:
See also this issue (which is more relevant to the specific situation we were facing in this repo prior to this change): https://github.com/dependabot/dependabot-core/issues/ 9682

PR #4445, opened 2024-06-23
Replace custom ErrorLogger with Rails error subscriber

davidrunger:
Docs: https://guides.rubyonrails.org/error_reporting.html

It seems like a good idea to follow/leverage the Rails standard way here, for the sake of convention with other Rails apps, if nothing else, and also to maybe make it somewhat easier to switch our error-tracking service from Rollbar to something else, if we ever want to do that.

However, there is some awkwardness (as flagged here re: Sentry https://github.com/rails/rails/pull/ 43625#issuecomment-1072514175) in determining whether an error-reporting framework (Rollbar, in our case) vs the Rails error subscriber should report a given error. To resolve this awkwardness, we will continue to have Rollbar automatically report _unhandled_ exceptions, because Rollbar has the advantage of also automatically adding a useful amount of additional context about the errors that the Rails error subscriber lacks. We will only send via Rollbar via the error subscriber any _handled_ (manually sent) exceptions that reach it.

The error subscriber will log all exceptions that reach it.


config/application.rb
```diff
@@ -64,6 +64,12 @@ class DavidRunger::Application < Rails::Application
   config.active_record.default_timezone = :utc
 
   config.middleware.insert_after(ActionDispatch::Static, Rack::Deflater) # gzip all responses
+  initializer(
+    'move Browser middleware after Rack::Attack',
+    after: 'rack-attack.middleware',
+  ) do |app|
+    app.middleware.move_after(Rack::Attack, Browser::Middleware)
+  end
```
davidrunger, line 72: (unrelated change; I wanted to bring all of the middleware-related config together)

PR #4457, opened 2024-06-24
Delete display.scss util stylesheet

davidrunger:
Tailwind will provide this same class (if needed), and we're not actually loading this stylesheet (anymore?), anyway.


davidrunger:
I mostly just want to use this build to see whether all of the test tasks would run. (They didn't, which is good.)

PR #4462, opened 2024-06-25
Include Toastify CSS styles in global styles.css stylesheet

davidrunger:
**Downside:** This will somewhat increase the weight of our styles.css stylesheet and will load these styles on some pages that don't need them. I think that this adds about 4 KiB of CSS to the styles.css stylesheet.

**Upside:** These styles can now be reused across more apps (just loaded once), which should theoretically make overall CSS served be less for users who visit multiple Toastify-using apps, and also these CSS imports won't clutter up our JavaScript imports or need to be repeated so many times in our JavaScript code (and it will be easier to use more toasters elsewhere in the future).

### DIFF
--- a/lib/test/tasks/run_file_size_checks.rb
+++ b/lib/test/tasks/run_file_size_checks.rb
@@ -20,7 +20,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'logs*.js' => (790..800),
     'marriage*.js' => (10..20),
     'quizzes*.js' => (123..133),
-    'styles*.css' => (5..15),
+    'styles*.css' => (9..19),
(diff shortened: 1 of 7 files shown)


lib/test/tasks/run_file_size_checks.rb
```diff
@@ -20,7 +20,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'logs*.js' => (790..800),
     'marriage*.js' => (10..20),
     'quizzes*.js' => (123..133),
-    'styles*.css' => (5..15),
+    'styles*.css' => (9..19),
```
davidrunger, line 23: I think it was about 10 KiB before (?), and now it's about 14 KiB.

PR #4514, opened 2024-07-01
[bin/lint/prettier] Don't print redundant success message

davidrunger:


davidrunger:
Closing in preference of #4515 (which also does RuboCop).

PR #4565, opened 2024-07-07
[pre-push] Remove TypeScript check

davidrunger:
Motivation: The TypeScript check seems to use up a pretty nontrivial amount of the not overly abundant resources on my little computer. And I think it's relatively rare that I push a PR that has TypeScript violations. So I'm not sure it's worth the cost (somewhat overwhelming my computer) to run the TypeScript checks locally. We will still check types in CI.


davidrunger:
I guess that a big part of the issue is that our other linters can run against only the relevant changed files, whereas the TypeScript check must run against the whole project. So I think it makes sense to leave this for CI only.

PR #4569, opened 2024-07-08
Add basic image/file uploading and viewing functionality [NEW-1]

davidrunger:


davidrunger:
I think that this is just not worth it. I think that what I am going to do instead is to upload files manually to S3. I think it might have been worth it if ActiveStorage would have played a little more nicely with my objective of uploading publicly accessible files, but it didn't, and so I was fighting the system. The `public: true` option in the S3 ActiveStorage config in `storage.yml` was causing uploads to fail for me. Instead of nice filenames like `david.jpg`, ActiveStorage just uses a key that is random alphanumeric characters without an extension. We'd have to add libvips installation to our GitHub Action and to the server box (assuming its not already on the server). We'd have to pull in some additional gem dependencies. We'd have to add another storage option. So, yeah, overall, I think just not worth it. But I'm glad I did the experiment, to uncover some of these issues; otherwise, I wouldn't have had much insight into whether or not it is worth it.

PR #4577, opened 2024-07-09
Add an emoji picker [NEW-13]

davidrunger:


tsconfig.json
```diff
@@ -1,21 +1,23 @@
...
-    "skipLibCheck": true,
     "allowJs": true,
+    "esModuleInterop": true,
+    "jsx": "preserve",
     "lib": ["esnext", "dom"],
-    "types": ["vite/client"],
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noImplicitAny": true,
+    "noUnusedLocals": true,
```
davidrunger, line 10: Mostly I'm just alphabetizing, but these `noImplicitAny` and `noUnuseLocals` properties are new.

package.json
```diff
@@ -25,9 +25,11 @@
     "dompurify": "^3.1.6",
     "easytimer.js": "^4.6.0",
     "element-plus": "^2.7.6",
+    "emojilib": "^3.0.12",
     "eslint": "^8.57.0",
     "eslint-import-resolver-custom-alias": "^1.3.2",
     "eslint-plugin-vue": "^9.27.0",
+    "fuse.js": "^7.0.0",
```
davidrunger, line 32: I'm pretty impressed with this library, both in terms of its performance and also the quality of its match results.

.eslintrc.js
```diff
@@ -33,6 +33,8 @@ module.exports = {
   },
   rules: {
     '@typescript-eslint/no-empty-function': 'off',
+    // https://github.com/vuejs/language-tools/issues/ 47#issuecomment-765562095
+    '@typescript-eslint/no-unused-vars': 'off',
```
davidrunger, line 37: ESLint complains about unused vars when using `script setup`. Instead, we'll leave it to TypeScript to warn us about unused vars; it seems not to complain about the "unused" variables in `script setup`. (I didn't test whether or not it complains about variables that actually do go unused, though.)

app/javascript/groceries/components/check_in_items_list.vue
```diff
@@ -4,7 +4,7 @@ section(v-if="items.length > 0")
 
   ul.check-in-items-list.mb-2
     li.flex.items-center.mb-2(
-      v-for='(item, index) in items'
+      v-for='item in items'
```
davidrunger, line 7: TypeScript [caught](https://github.com/davidrunger/david_runger/actions/runs/9865525548/job/27242625078?pr=4577#step:10:239) some unused variables (even in templates; in fact, exclusively in templates, I guess maybe because ESLint wasn't going there before), presumably because we enabled the `noUnusedLocals` TypeScript compiler option.

PR #4604, opened 2024-07-11
[groceries/check-in] Prevent horizontal scroll [GROC-9]

davidrunger:


app/javascript/groceries/components/check_in_items_list.vue
```diff
@@ -14,9 +14,10 @@ section(v-if="items.length > 0")
         :id='`trip-checkin-item-${item.id}`'
       )
       label.ml-2(:for='`trip-checkin-item-${item.id}`')
-        span.mr-1(:class='{ "text-gray-500": item.skipped }')
```
davidrunger, line 17: This `mr-1` was causing the check-in content to be slightly horizontally scrollable when the right-margin was dangling at the end of the line and pushing the width of that `li` to be a bit wider than its container.

app/javascript/groceries/components/check_in_items_list.vue
```diff
@@ -3,7 +3,7 @@ section(v-if="items.length > 0")
   h3.font-bold.mb-2 {{ title }} ({{ items.length }})
 
   ul.check-in-items-list.text-base.mb-2
-    li.flex.items-center.mb-2(
+    li.flex.items-center.break-word.mb-2(
```
davidrunger, line 6: This makes sure that long words break (on word boundaries) rather than making the `li` indefinitely wide if the element contains length text.

app/javascript/groceries/components/check_in_items_list.vue
```diff
@@ -14,9 +14,10 @@ section(v-if="items.length > 0")
         :id='`trip-checkin-item-${item.id}`'
       )
       label.ml-2(:for='`trip-checkin-item-${item.id}`')
-        span.mr-1(:class='{ "text-gray-500": item.skipped }')
+        span(:class='{ "text-gray-500": item.skipped }')
           span {{item.name}}
           span(v-if='item.needed > 1') {{' '}} ({{item.needed}})
+        span {{ ' ' }}
```
davidrunger, line 20: I'm adding this space in place of the `mr-1`. I don't love this solution because it means that the only amount of spacing we can have here is the width of a space. In this case, that's just about what we want, but I don't like the feeling that it might not have been quite what we want, and I would struggle to come up with a good solution in such a case. The problem is that we want the spacer to be of width X, _unless_ the "Skip" button breaks to begin its own new line, in which case we _don't_ want the spacer to take up with at the beginning of that line, effectively indenting the "Skip" button.

app/javascript/groceries/components/check_in_items_list.vue
```diff
@@ -68,7 +69,8 @@ export default defineComponent({
 </script>
 
 <style lang="scss">
-.el-button.is-link {
+// double class in selector to increase specificity of this override
+.el-button.el-button.is-link {
```
davidrunger, line 73: I noticed that, at least in my development environment, sometimes this style wasn't overriding the Element Plus CSS, as we intend for it to do.

PR #4637, opened 2024-07-19
Downgrade tailwindcss from 3.4.6 to 3.4.4

davidrunger:
A bug seems to have been introduced, I think in 3.4.5, which is currently causing some visual regressions in production. The problem seems to be that, in some cases, when using pug templates, the last listed class does not get picked up (for example, a `z-10` class in Modal.vue and `leading-unset` in Item.vue).

Possibly relevant issues:

https://github.com/tailwindlabs/tailwindcss/issues/ 14005

https://github.com/tailwindlabs/tailwindcss/issues/ 14013

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "stylelint-config-standard-scss": "^11.1.0",
     "stylelint-scss": "^5.3.2",
     "tailwind-config-viewer": "^2.0.4",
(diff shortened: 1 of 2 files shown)


davidrunger:
The Percy snapshot shows a visual regression being fixed on the homepage.

PR #4641, opened 2024-07-19
[logs/text] Show a delete button next to each entry [LOG-10]

davidrunger:


app/models/user.rb
```diff
@@ -33,6 +33,7 @@ class User < ApplicationRecord
   has_many :requests, dependent: :destroy
   has_many :stores, dependent: :destroy
   has_many :items, through: :stores # must come after has_many :stores declaration
+  has_many :number_log_entries, through: :logs
```
davidrunger, line 36: I don't think we ended up using this, but I think we might as well add it, for symmetry with `text_log_entries`.

PR #4679, opened 2024-07-21
Use freezolite (no more frozen_string_literal comments) [DEV-2]

davidrunger:
One downside of this is that, because we no longer have explicit `frozen_string_literal: true` comments at the top of files, RuboCop does not know that strings are frozen by default, and so RuboCop's autocorrections (applied in this change) include adding explicit `.freeze` calls to string constants. This is not totally ideal, but I think it's okay.

I wonder if maybe I am going to miss having these comments at the top of my files. I sort of felt like they provided a pleasant bit of "breathing room" at the top of the file, rather than smashing the top of the class definition in a file up against the top of my editor.

I guess let's try it and see how it feels.


.rubocop.yml
```diff
@@ -16,3 +16,6 @@ AllCops:
   <% `git status --ignored --porcelain`.lines.grep(/^!! /).each do |path| %>
     - <%= path.sub(/^!! /, '').sub(/\/$/, '/**/*').rstrip %>
   <% end %>
+
+Style/FrozenStringLiteralComment:
+  EnforcedStyle: never
```
davidrunger, line 21: Here we are overriding the `runger_style` rule (rather than changing it upstream) because most projects that use `runger_style` aren't also going to use `freezolite` and so they should continue to use the `frozen_string_literal: true` comments.

davidrunger:
> I wonder if maybe I am going to miss having these comments at the top of my files. I sort of felt like they provided a pleasant bit of "breathing room" at the top of the file, rather than smashing the top of the class definition in a file up against the top of my editor.

I found that this can sort of be addressed via `"editor.padding.top": 10` in VS Code.

PR #4682, opened 2024-07-21
Provide a copy-to-clipboard button for the quiz link [QUIZ-1]

davidrunger:


app/views/quizzes/show.html.haml
```diff
@@ -1,4 +1,5 @@
 - content_for(:page_assets) do
+  = ts_tag('copy_to_clipboard')
```
davidrunger, line 2: I created a new, separate pack to hold this `copy_to_clipboard` functionality because I can imagine it being useful in other contexts where we want to be able to copy-to-clipboardify Rails-rendered buttons (as in this case). So, it will be nice to just be able to add this `ts_tag('copy_to_clipboard')` call wherever we want to add that functionality.

PR #4694, opened 2024-07-23
Add a dedicated proposal page

davidrunger:
GROC~3 calls for us to link to a place where users can invite their partner to the app as their partner. The only place that we currently have to do that has a URL of `/check_ins`. We want to add a dedicated place with a more appropriate URL as a generic place to make a proposal.


Gemfile.lock
```diff
@@ -563,7 +563,7 @@ GEM
     runger_config (5.2.0)
       activesupport (>= 7.1.2)
     runger_email_reply_trimmer (0.3.0)
-    runger_style (2.12.0)
+    runger_style (2.13.0)
```
davidrunger, line 566: I pulled this update in here so that I can have an example description line that is more than 100 characters.

PR #4699, opened 2024-07-24
Extract CheckInModal component from Store [GROC-15]

davidrunger:


app/javascript/groceries/components/Store.vue
```diff
@@ -315,9 +258,4 @@ button.choose-stores {
     color: black;
   }
 }
-
-.check-in-items-list {
-  max-height: 50vh;
-  overflow: auto;
-}
```
davidrunger, line 322: Deleting this is unrelated, but it's something that I noticed while looking over this CSS to see if I should transfer any of it to the new component. The relevant element with this class was moved to a different component, so this rule will never have any effect.

davidrunger: As an aside, I think that this illustrates one advantage of Tailwind, which is that the styles are tightly coupled to the HTML. So, when I extract/move one bit of template from one component to another, with Tailwind, the styles move along automatically. In contrast, when using CSS, then the CSS can be "left behind", as happened here.

PR #4719, opened 2024-07-25
Live update LogEntry on edit and delete [LOG-5]

davidrunger:


app/javascript/logs/store.ts
```diff
@@ -131,7 +131,7 @@ export const useLogsStore = defineStore('logs', {
       logEntry: LogEntry;
     }) {
       log.log_entries = log.log_entries.filter(
-        (logEntry) => logEntry !== logEntryToDelete,
+        (logEntry) => logEntry.id !== logEntryToDelete.id,
```
davidrunger, line 134: We need to compare by id now, because the object that we receive via the websocket update will never be the same object as any of the existing objects in the store.

PR #4751, opened 2024-07-31
[Dockerfile] Remove RAILS_MASTER_KEY build arg

davidrunger:
Motivation: make the build process slightly simpler.

Instead of having this as a build arg, we'll expect the production RAILS_MASTER_KEY to be present in a `.env` file on the machine doing the building.

Also, just put DOCKER_BUILD and GIT_REV as env variable prefixes for the assets:precompile command that needs them, rather than adding them, in general, to the ENV.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,14 +55,11 @@ RUN bundle exec bootsnap precompile --gemfile
 COPY . .
 
 ARG GIT_REV
-ARG RAILS_MASTER_KEY
-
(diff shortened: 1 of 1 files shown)


davidrunger:
I'm going to close this and do this work in another branch with a different name.

davidrunger:
Here's the new PR: https://github.com/davidrunger/david_runger/pull/4753 .

PR #4761, opened 2024-07-31
[bootstrap-letsencrypt] Use --test-cert for staging (not --staging)

davidrunger:
https://letsencrypt.org/docs/staging-environment/ :

> If you’re using Certbot, you can use our staging environment with the `--test-cert` or `--dry-run` flag.

### DIFF
--- a/bin/ssl/bootstrap-letsencrypt.sh
+++ b/bin/ssl/bootstrap-letsencrypt.sh
@@ -64,7 +64,7 @@ case "$email" in
 esac
 
 # Enable staging mode if needed
-if [ $staging != "0" ]; then staging_arg="--staging"; fi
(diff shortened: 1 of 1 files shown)


davidrunger:
Actually, I guess that `--staging` works, too: https://eff-certbot.readthedocs.io/en/stable/using.html

```
  --test-cert, --staging
                        Use the Let's Encrypt staging server to obtain or
                        revoke test (invalid) certificates; equivalent to
                        --server https://acme-
                        staging-v02.api.letsencrypt.org/directory (default:
                        False)
```

davidrunger:
Back again on the other hand, maybe not. `docker compose run --rm --entrypoint "certbot --help" certbot` gives:

```
  --test-cert       Obtain a test certificate from a staging server
```

So it does seem like `--test-cert` is the safer option.

PR #4800, opened 2024-08-03
Copy deploy script to server for execution

davidrunger:
There seems to be an issue with our current approach where `docker compose exec` swallows all the rest of the script. Hopefully this will avoid that.


davidrunger:
Based on my testing, I'm pretty sure that this will indeed fix the issue. But there's only one way to find out for sure... deploy!

PR #4805, opened 2024-08-03
[deploy] Run the server install script as part of each deploy

davidrunger:
Usually, this will have no effect (since the install script will usually have already been executed, and since it should be idempotent), but running the install script each time that we deploy makes it easy to ensure that the latest version is always "installed". Since the install script should be fast and idempotent, there shouldn't really be much downside to doing this.

### DIFF
--- a/bin/server/boot-services.sh
+++ b/bin/server/boot-services.sh
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# Boot the services that we want/need booted in order to run the web app.
+
 set -euo pipefail # exit on any error, don't allow undefined variables, pipes don't swallow errors
 
+# NOTE: These aren't the only services that need to be booted, but, via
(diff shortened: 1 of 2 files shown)


bin/server/boot-services.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# Boot the services that we want/need booted in order to run the web app.
```
davidrunger, line 3: (I'm just adding some unrelated comments/documentation to this file.)

bin/server/deploy.sh
```diff
@@ -13,6 +13,9 @@ if [ "$current_commit" != "$GIT_REV" ] ; then
   exit 1
 fi
 
+# Run the install script.
+bin/server/install.sh
```
davidrunger, line 17: Here's the main change in this PR.

PR #4810, opened 2024-08-04
Allow users to delete their account

davidrunger:
I think that this is an important feature to provide to users for privacy / data control reasons. Also, I'm working on a privacy policy, and I want to be able to say in the privacy policy that users can delete their accounts via the web UI.


app/views/marriages/show.html.haml
```diff
@@ -1,5 +1,5 @@
 - content_for(:page_assets) do
-  = ts_tag('marriage')
+  = ts_tag('ujs')
```
davidrunger, line 2: This pack didn't actually have anything marriage-specific in it. It just imports and initializes Rails UJS. So, I renamed it to `ujs`, so that it can be included elsewhere (as done in this PR) without confusion.

app/views/users/edit.html.haml
```diff
@@ -1,22 +1,31 @@
...
-          %span
-            = form.label :secret, 'Secret (required)'
-            = form.text_field :secret
+- content_for(:page_assets) do
+  = ts_tag('ujs')
 
-          %span.ml-8
-            = form.label :name, 'Name (optional)'
-            = form.text_field :name
+%div.my-8
```
davidrunger, line 4: I suggest viewing the diff with whitespace changes ignored: https://github.com/davidrunger/david_runger/pull/4810/files?w=1 .

spec/features/vue_playground_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe 'Vue Playground', :prerendering_disabled do
```
davidrunger, line 1: The specific examples that need prerendering disabled have it marked below. We don't need to mark it up here, and, by removing this, there is now one spec we run with prerendering enabled where it wasn't before, which I think is a small win (diverging less from production).

spec/spec_helper.rb
```diff
@@ -239,6 +239,19 @@
...
+    original_enable = Bullet.enable?
+    original_counter_cache_enable = Bullet.counter_cache_enable?
+
+    Bullet.enable = false
+    Bullet.counter_cache_enable = false
+
+    example.run
+  ensure
+    Bullet.enable = original_enable
+    Bullet.counter_cache_enable = original_counter_cache_enable
```
davidrunger, line 252: See this commit message for discussion of spec flakiness that is introduced if we don't include this line of code: https://github.com/davidrunger/david_runger/pull/4810/commits/4489932347650c325519219214aa7817735ce534

PR #4815, opened 2024-08-04
[docker] Make RAILS_ENV a build arg [DEV-70]

davidrunger:
This way, we can build with `RAILS_ENV=production` for production and `RAILS_ENV=development` when running locally.

However, even when we boot the app locally with `RAILS_ENV=development`, we still don't install gems from the `development` or `test` bundler groups. So, this PR adds a number of additional checks against a new `IS_DOCKER` constant, so that we don't try to use development-/test- only gems when running the app locally via Docker (with `RAILS_ENV=development`).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,5 @@
 .env*
+!.env.development.local
 !.env.production.local
 .git
 .github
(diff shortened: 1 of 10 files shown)


davidrunger:
The app can be accessed, when booted locally via Docker in development mode, at https://localhost/ .

PR #4820, opened 2024-08-04
[docker-compose] Rework depends_on listings for Rails services

davidrunger:
The problem with what we were doing before is that we had some service-specific `depends_on` declarations that were overriding (as opposed to adding to) the default `depends_on` declaration for `x-rails-config`. For example, `clock` was not depending on `redis`, which it should do.

I think that this might be the cause of errors like this when trying to boot `clock`:

> clock-1 | Skedjewel 0.0.13 is running with PID 1. > clock-1 | Unhandled exception: Socket::ConnectError: Error connecting to 'localhost:6379': Connection refused (Redis::CannotConnectError)

This new configuration has most Rails services `depends_on` initialize_database. The exception is initialize_database itself, which depends on memcached, postgres, and redis. Hopefully this setup will ensure a reliable bootup chain.

### DIFF
--- a/bin/server/deploy.sh
+++ b/bin/server/deploy.sh
@@ -24,8 +24,8 @@ bin/server/install.sh
 # Rebuild the app.
 bin/build-docker production
 
-# Launch new Rails services.
-docker compose up --detach clock nginx web worker
+# Launch fresh services.
+docker compose up --detach
 
 # Run release tasks.
 docker compose exec web bin/server/release-tasks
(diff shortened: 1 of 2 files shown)


bin/server/deploy.sh
```diff
@@ -24,8 +24,8 @@ bin/server/install.sh
 # Rebuild the app.
 bin/build-docker production
 
-# Launch new Rails services.
-docker compose up --detach clock nginx web worker
+# Launch fresh services.
+docker compose up --detach
```
davidrunger, line 28: Sort of an unrelated change, but I think it's maybe better / more reliable to just relaunch everything. I can imagine that this might possible make resolution of the `depends_on` chains more reliable.

PR #4823, opened 2024-08-05
[puma] Add set_remote_address(header: 'X-Forwarded-For')

davidrunger:
I think that this _might_ make `request.ip` resolve to the originating user's IP address, rather than the Cloudflare address, as is currently happening. Let's try it and see.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -23,6 +23,10 @@ environment(ENV.fetch('RAILS_ENV', 'development'))
 #
 workers(ENV.fetch('WEB_CONCURRENCY', 0))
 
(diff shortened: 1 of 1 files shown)


davidrunger:
This seems to have worked. I'm not sure why we didn't need to do this before, though? 🤔 🤷‍♂️

davidrunger:
It's interesting that our `Request`s were logging the end-user's IP (never a Cloudflare IP) even before making this change. So, I guess that whatever the problem was only affected the Rack Attack IP address detection, and not Rails's IP address detection.

davidrunger:
This fixes https://davidrunger.atlassian.net/browse/DEV-77 .

PR #4826, opened 2024-08-05
[nginx] Remove 'Forwarded' header

davidrunger:
We noticed that rack-attack was blocking the Cloudflare IP that forwarded the bad request, rather than the originating client IP.

I figured out a fix for this: #4823. However, I have been wondering why that fix was even necessary.

I suspect that it might have been necessary because of the fact that Rack [prioritizes][1] the `Forwarded` header over `X-Forwarded-For`. I suspect that maybe the way that these headers are handled (maybe as they relate to `cloudflare-rails`?) is different, and/or maybe we aren't formatting the `Forwarded` header correctly. Maybe on Dokku (before our recent migration to raw Docker (Compose)) we weren't sending a `Forwarded` header at all, and that's why we didn't need #4823 when on Dokku.

[1]: https://github.com/rack/rack/blob/v3.1.7/lib/rack/request.rb#L43

Therefore, I am going to try reverting the #4823 change and also stop sending a `Forwarded` header. If, after this change, rack-attack continues to block the originating client IP rather than the Cloudflare IP, then my above theory is probably correct, and we'll leave things as they will at that point be. If we see a regression, then my theory is probably not correct, and we'll probably revert this PR. But I would like to try it and see what happens.

Unfortunately, it's sort of necessary to test this in production, because outside of a production environment it is difficult to simulate what the real request headers and environment etc will be when these requests are handled.

One nice thing about this change is that it allows us to delete a decent amount of NGINX code.

### DIFF
--- a/config/nginx/nginx.conf
+++ b/config/nginx/nginx.conf
@@ -53,27 +53,6 @@ http {
         ""      close;
     }
 
-    map $remote_addr $proxy_forwarded_elem {
-
-        # IPv4 addresses can be sent as-is
-        ~^[0-9.]+$        "for=$remote_addr";
-
-        # IPv6 addresses need to be bracketed and quoted
-        ~^[0-9A-Fa-f:.]+$ "for=\"[$remote_addr]\"";
-
-        # Unix domain socket names cannot be represented in RFC 7239 syntax
-        default           "for=unknown";
-    }
-
-    map $http_forwarded $proxy_add_forwarded {
-
-        # If the incoming Forwarded header is syntactically valid, append to it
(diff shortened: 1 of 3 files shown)


davidrunger:
This fixes [https://davidrunger.atlassian.net/browse/](https://davidrunger.atlassian.net/browse/DEV-77)[DEV-77](https://davidrunger.atlassian.net/browse/DEV-77) .

[DEV-77]: https://davidrunger.atlassian.net/browse/DEV-77?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4827, opened 2024-08-05
[deploy] Add '--remove-orphans' to 'docker compose up' command

davidrunger:
This should hopefully prevent disk usage from growing whenever we deploy, and maybe eventually we will be able to do away with the `docker system prune --all --force` cron job.

The reason I initially hesitated to do this is because I didn't want to waste time/CPU deleting orphans when trying to boot up new containers. However, from using the `--remove-orphans` option locally, it doesn't feel like it adds a lot of time, and so I think it might be worth doing.

### DIFF
--- a/bin/server/deploy.sh
+++ b/bin/server/deploy.sh
@@ -25,7 +25,7 @@ bin/server/install.sh
 bin/build-docker production
 
 # Launch fresh services.
-docker compose up --detach
+docker compose up --detach --remove-orphans
 
 # Run release tasks.
 docker compose exec web bin/server/release-tasks


davidrunger:
I think that the build cache will continue to grow indefinitely, and this change will probably only shrink the growth rate of image and/or container disk usage. Still, I think that will be a help. We could maybe decrease the `docker system prune --all --force` frequency even more from weekly to monthly or something, but maybe we don't want to do that, anyway, since it's probably good to build fresh images on a ~weekly basis, to pull in the latest system package updates and whatnot.

PR #4869, opened 2024-08-09
Serve the blog directly from Rails (not Cloudflare Pages) [BLOG-10]

davidrunger:
The blog side of this has been done here: https://github.com/davidrunger/blog/pull/277 .


spec/config/initializers/request_logging_spec.rb
```diff
@@ -5,8 +5,8 @@
         # rubocop:disable RSpec/AnyInstance
         allow_any_instance_of(ActionController::TestRequest).
           to receive(:headers).
-          and_wrap_original do |request|
-            headers = request.call
+          and_wrap_original do |original_headers_method|
+            headers = original_headers_method.call
```
davidrunger, line 9: (Unrelated change; I just noticed that the block parameter name here is confusing; I am guessing it's some sort of copy-paste carelessness.)

spec/spec_helper.rb
```diff
@@ -284,8 +284,8 @@
 
     allow_any_instance_of(ActionController::TestRequest).
       to receive(:headers).
-      and_wrap_original do |request|
-        headers = request.call
+      and_wrap_original do |original_headers_method|
+        headers = original_headers_method.call
```
davidrunger, line 288: (Unrelated change; I just noticed that the block parameter name here is confusing; I am guessing it's some sort of copy-paste carelessness.)

PR #4871, opened 2024-08-09
Delete 500.html error page

davidrunger:
Instead, we now use `app/views/errors/internal_error.html.erb`.


config/environments/test.rb
```diff
@@ -32,11 +32,14 @@
 
   # Show full error reports and disable caching.
   config.consider_all_requests_local = true
-  config.action_controller.perform_caching = false
-  config.cache_store = :null_store
+  # Render exceptions via ErrorsController
+  config.exceptions_app = routes
```
davidrunger, line 36: This reflects a configuration that we also have in production, and it's necessary in order for the `spec/features/error_handling_spec.rb` tests to pass.

config/environments/test.rb
```diff
@@ -32,11 +32,14 @@
...
 
   config.action_dispatch.show_exceptions = :none
 
+  config.action_controller.perform_caching = false
+  config.cache_store = :null_store
```
davidrunger, line 41: I'm just moving these so that the error configs are grouped together.

PR #4886, opened 2024-08-11
Switch back to 'config.public_file_server.enabled = true'

davidrunger:
I think I was wrong to think that NGINX server our `public/` directory. (Maybe it should, but I think it currently doesn't? At least, that's what the current errors in production suggest.)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -62,6 +62,7 @@ class DavidRunger::Application < Rails::Application
     g.factory_bot(dir: 'spec/factories')
   end
 
+  config.middleware.insert_after(ActionDispatch::Static, Rack::Deflater) # gzip all responses
   initializer(
     'move Browser middleware after Rack::Attack',
     after: 'rack-attack.middleware',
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -28,7 +28,7 @@
   config.require_master_key = !IS_DOCKER_BUILD
 
   # Disable serving static files from `public/`, relying on NGINX/Apache to do so instead.
-  config.public_file_server.enabled = false
```
davidrunger, line 31: We changed this to `false` here https://github.com/davidrunger/david_runger/pull/4884/files#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eR31 but probably shouldn't have.

config/application.rb
```diff
@@ -62,6 +62,7 @@ class DavidRunger::Application < Rails::Application
     g.factory_bot(dir: 'spec/factories')
   end
 
+  config.middleware.insert_after(ActionDispatch::Static, Rack::Deflater) # gzip all responses
```
davidrunger, line 65: This undoes https://github.com/davidrunger/david_runger/pull/4885 .

PR #4889, opened 2024-08-11
Retry 'config.public_file_server.enabled = false' [DEV-90]

davidrunger:
... now with the NGINX changes such that it will hopefully work.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -62,7 +62,6 @@ class DavidRunger::Application < Rails::Application
     g.factory_bot(dir: 'spec/factories')
   end
 
-  config.middleware.insert_after(ActionDispatch::Static, Rack::Deflater) # gzip all responses
   initializer(
     'move Browser middleware after Rack::Attack',
     after: 'rack-attack.middleware',
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -28,7 +28,7 @@ Rails.application.configure do
   config.require_master_key = !IS_DOCKER_BUILD
 
   # Disable serving static files from `public/`, relying on NGINX/Apache to do so instead.
-  config.public_file_server.enabled = true
+  config.public_file_server.enabled = false
 
   # Compress CSS using a preprocessor.
   # config.assets.css_compressor = :sass
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,9 +57,12 @@ services:
     profiles:
       - nondefault
     volumes:
-      - ./config/nginx/:/etc/nginx/
+      - ./config/nginx/nginx.conf:/etc/nginx/nginx.conf:ro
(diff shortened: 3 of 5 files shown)


config/application.rb
```diff
@@ -62,7 +62,6 @@ class DavidRunger::Application < Rails::Application
     g.factory_bot(dir: 'spec/factories')
   end
 
-  config.middleware.insert_after(ActionDispatch::Static, Rack::Deflater) # gzip all responses
```
davidrunger, line 65: There's no longer a need for this, since NGINX will compress the assets with gzip. It also will no longer work, since Rails won't add the `ActionDispatch::Static` middleware.

config/environments/production.rb
```diff
@@ -28,7 +28,7 @@
   config.require_master_key = !IS_DOCKER_BUILD
 
   # Disable serving static files from `public/`, relying on NGINX/Apache to do so instead.
-  config.public_file_server.enabled = true
+  config.public_file_server.enabled = false
```
davidrunger, line 31: This says "don't use Rails to serve `public/` files" (since NGINX will do it, instead).

docker-compose.yml
```diff
@@ -57,9 +57,12 @@ services:
     profiles:
       - nondefault
     volumes:
-      - ./config/nginx/:/etc/nginx/
+      - ./config/nginx/nginx.conf:/etc/nginx/nginx.conf:ro
+      - ./config/nginx/nginxconfig.io/:/etc/nginx/nginxconfig.io/:ro
+      - ./config/nginx/sites-enabled/:/etc/nginx/sites-enabled/:ro
```
davidrunger, line 62: Enumerating specific files/directories allows us to do this above: `include /etc/nginx/mime.types;`. Previously, we had been overwriting that file, by overwriting the whole `/etc/nginx/` directory with our own config (which doesn't include `mime.types`).

davidrunger: (`...:ro` means "read only".)

docker-compose.yml
```diff
@@ -57,9 +57,12 @@ services:
...
       - ./ssl-data/certbot/conf:/etc/letsencrypt
       - ./ssl-data/certbot/www:/var/www/_letsencrypt
+      - app-public:/var/www/david-runger-public:ro
```
davidrunger, line 65: Here we link `public/` from the `web` service to `/var/www/david-runger-public` in the `nginx` service.

docker-compose.yml
davidrunger: > Enumerating specific files/directories allows us to do this above: `include /etc/nginx/mime.types;`. Previously, we had been overwriting that file, by overwriting the whole `/etc/nginx/` directory with our own config (which doesn't include `mime.types`).

I think that this is wrong; see https://github.com/davidrunger/david_runger/pull/4890 .

PR #4906, opened 2024-08-14
[nginx] Write access logs every second [DEV-84]

davidrunger:
This will be helpful (vs our previous 1 minute buffer) when debugging issues in real time / live tailing logs.

### DIFF
--- a/config/nginx/sites-enabled/davidrunger.com.conf
+++ b/config/nginx/sites-enabled/davidrunger.com.conf
@@ -13,7 +13,7 @@ server {
     include                 nginxconfig.io/security.conf;
 
     # logging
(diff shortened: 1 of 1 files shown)


config/nginx/sites-enabled/davidrunger.com.conf
```diff
@@ -13,7 +13,7 @@ server {
     include                 nginxconfig.io/security.conf;
 
     # logging
-    access_log              /var/log/nginx/access.log combined buffer=512k flush=1m;
+    access_log              /var/log/nginx/access.log combined buffer=4k flush=1s;
```
davidrunger, line 16: I think that we will flush logs whenever we meet either the 4k condition or the 1s condition. I think it will be much more common for us to reach the 1s condition first.

PR #4938, opened 2024-08-17
[docker] Don't send development logs to Papertrail

davidrunger:
In development, developers can now avoid sending logs to Papertrail (and instead just have logs available locally, via the `local` driver) via `cp docker-compose.override.yml.example docker-compose.override.yml`. The `docker-compose.override.yml` is a special file name that is loaded automatically by Docker.

The reason that we are using a `docker-compose.override.yml.example` template (and gitignoring the actual `docker-compose.override.yml` file) is because we _do_ want to track at least a copy of this file via version control, but we _don't_ want this file to be sent to the server via git, because, on the server, we _do_ want logs to be sent to Papertrail.

Also, instead of using `.env.papertrail.local` to source `PAPERTRAIL_URL`, we will now use `.env`, which is sourced by Docker Compose automatically, which makes things a lot easier and which is a more complete/foolproof solution. For example, even the code being deleted in this PR that loads `.env.papertrail.local` manually in several cases doesn't help when e.g. running `docker compose run --rm web echo 'hi'` on the server; we'd still get warnings like this:

> WARN[0000] The "PAPERTRAIL_URL" variable is not set. Defaulting to a blank string.

Just using `.env` (on the server) should avoid these warnings in all cases.

Also, in order to avoid warnings in development if a `.env` file with `PAPERTRAIL_URL` is not present, we also now provide a default value for `PAPERTRAIL_URL` in `docker-compose.yml`.

### DIFF
--- a/docker-compose.override.yml.example
+++ b/docker-compose.override.yml.example
@@ -1,0 +1,13 @@
+x-rails-json-file-logging: &json-file-logging
+  logging:
+    driver: local
+
+services:
+  clock:
+    <<: *json-file-logging
+  initialize_database:
+    <<: *json-file-logging
+  web:
+    <<: *json-file-logging
+  worker:
+    <<: *json-file-logging
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@
 .byebug_history
 .irb_history
 
+# Docker
+docker-compose.override.yml
+
 # Generated JavaScript path helpers
 /app/javascript/rails_assets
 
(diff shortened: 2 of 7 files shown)


docker-compose.override.yml.example
```diff
@@ -0,0 +1,13 @@
...
+
+services:
+  clock:
+    <<: *json-file-logging
+  initialize_database:
+    <<: *json-file-logging
+  web:
+    <<: *json-file-logging
+  worker:
+    <<: *json-file-logging
```
davidrunger, line 13: Unfortunately, this list of services that use the `x-rails-config` of `docker-compose.yml` is WET, and I cannot really think of a good / reasonably simple way to avoid that. Not the end of the world, I think.

PR #4950, opened 2024-08-18
[docker-compose] Add health check for the web service [DEV-93]

davidrunger:


Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN --mount=type=cache,sharing=private,target=/var/lib/apt/lists \
   --mount=type=cache,sharing=private,target=/var/cache/apt \
   apt-get update && \
   apt-get install --no-install-recommends -y \
-  libjemalloc2 postgresql-client
+  curl libjemalloc2 postgresql-client
```
davidrunger, line 16: It's kind of sad to be adding `curl` to the image just for the sake of the health check, but it's not the end of the world.

app/controllers/health_checks_controller.rb
```diff
@@ -0,0 +1,25 @@
+class HealthChecksController < ApplicationController
+  skip_before_action :authenticate_user!, only: :index
+  before_action :skip_authorization, only: :index
+
+  def index
+    [
+      Thread.new { check_postgres },
+      Thread.new { check_redis },
+    ].each(&:join)
```
davidrunger, line 9: Parallelize these for maximum speed.

app/controllers/health_checks_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    head :ok
+  end
+
+  private
+
+  def check_postgres
+    ActiveRecord::Base.logger.silence do
```
davidrunger: This is particularly so that we don't see ActiveRecord query logs in development.

config/initializers/lograge.rb
```diff
@@ -6,4 +6,5 @@
   config.lograge.enabled = true
   config.lograge.custom_options = ->(event) { Logs::LogBuilder.new(event).extra_logged_data }
   config.lograge.formatter = ->(data) { Logs::LogFormatter.new(data).call }
+  config.lograge.ignore_actions = ['HealthChecksController#index']
```
davidrunger, line 9: We want to suppress these logs because this action will be requested a lot, and we don't want that much (mostly low value) clutter in our logs.

PR #4960, opened 2024-08-18
[sitemap] Add html suffix for privacy policy

davidrunger:
It's needed (maybe since moving to serve static assets via NGINX?).

I'm not sure there's a lot of advantage in listing the privacy policy in the sitemap at all, but I guess it seems sort of like a good idea. Maybe it enhances credibility and makes search engines rank me higher.


davidrunger:
Closing in favor of #4961.

PR #4968, opened 2024-08-19
[workout] Prevent screen from locking if workout in progress [WORK-1]

davidrunger:


lib/test/tasks/run_file_size_checks.rb
```diff
@@ -15,7 +15,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'groceries*.css' => (93..103),
     'groceries*.js' => (410..420),
     'home*.css' => (0..10),
-    'home*.js' => (136..146),
+    'home*.js' => (141..151),
```
davidrunger, line 18: It's kind of disappointing that home.js is increasing (I guess because we also use vueuse there?).

PR #4969, opened 2024-08-20
[groceries] Show "Undo" button in toast after deleting item [GROC-5]

davidrunger:


app/models/item.rb
```diff
@@ -27,6 +27,7 @@ class Item < ApplicationRecord
   has_paper_trail
 
   scope :needed, -> { where('items.needed > 0') }
+  scope :unneeded, -> { where(needed: 0) }
```
davidrunger, line 30: (Only used in a spec right now, but I'm okay with that.)

spec/features/groceries_spec.rb
```diff
@@ -9,16 +9,21 @@
...
         expect(page).to have_button('Check in items')
 
         needed_item = store.items.needed.first!
+        unneeded_item = store.items.unneeded.first!
         expect(page).to have_text(/#{needed_item.name} +\(#{needed_item.needed}\)/)
 
+        # NOTE: When running specs with the development vite server, the
+        # following line triggers two warnings, which I'm pretty sure are caused
+        # by a bug in cuprite (namely that focus and blur events are of type
+        # `Event` rather than `FocusEvent`).
```
davidrunger, line 26: The bug is fixed by changing this line of code: https://github.com/rubycdp/cuprite/blob/v0.15.1/lib/capybara/cuprite/javascripts/index.js#L419

from this

```js
event = this.obtainEvent(name);
```

to this

```js
event = document.createEvent("FocusEvent");
event.initEvent(name, true, true)
```

spec/features/groceries_spec.rb
```diff
@@ -9,16 +9,21 @@
 
       let(:new_item_name) { 'blueberries' }
 
-      it 'allows adding an item' do
+      it 'allows adding an item, deleting an item, undoing the deletion, and checking in a shopping trip' do
```
davidrunger, line 12: (The "checking in a shopping trip" specs had been added previously, but they weren't added to this example description, which they should have been. Doing so now.)

app/policies/reifications_policy.rb
```diff
@@ -0,0 +1,11 @@
+class ReificationsPolicy < ApplicationPolicy
+  def create?
+    own_record?
+  end
+
+  private
+
+  def own_record?
+    @record.reify.user == @user
```
davidrunger, line 9: (We can't do `@record.reify.user_id == @user.id` here because the user might be a "through" association, i.e. there might not be a `user_id` column on the model in question.)

app/javascript/shared/http.ts
```diff
@@ -33,11 +33,18 @@ if (csrfToken) {
 }
 
 const kyApi = ky.extend({
+  headers: {
+    'Content-Type': 'application/json',
+  },
```
davidrunger, line 39: Ensure that all of these requests are made as JSON requests.

I think this actually isn't super necessary, because I think that all API routes are automatically considered JSON, anyway (per https://github.com/davidrunger/david_runger/blob/9ddd20b503420b9ca59c0834e2709b4426fbada4/config/routes.rb#L69), but it's still probably a good idea to put this here, even if it's redundant, for now.

app/javascript/shared/http.ts
```diff
@@ -33,11 +33,18 @@ if (csrfToken) {
...
   hooks,
   throwHttpErrors: false,
 });
 
 export const http = {
+  async delete<T>(url: string) {
```
davidrunger, line 45: I chose not to include a payload in the method signature. We could add that, in the future, if needed. However, I think it's uncommon for DELETE requests to have a payload, so I think we might never need it.

app/javascript/packs/logs_app.ts
```diff
@@ -8,11 +8,15 @@ import { renderApp } from '@/shared/customized_vue';
 import { useElementPlus } from '@/shared/element_plus';
 
 const app = renderApp(LogApp);
+
```
davidrunger, line 11: (I initially modified this file substantively, but ended up reverting those changes. I'm leaving these cosmetic changes.)

app/javascript/lib/toasts.ts
```diff
@@ -8,16 +8,16 @@ const TYPE_TO_CLASS_MAP = {
...
+  type?: ToastType;
+};
+
+export function toast(message: string, options?: ToastifyOptions) {
   Toastify({
     text: message,
     position: 'center',
     duration: 3000,
     className: options?.type && TYPE_TO_CLASS_MAP[options?.type],
-    escapeMarkup: !options?.html,
```
davidrunger, line 20: I'm removing the option to render HTML toasts via toastify, because all current such use cases are now being handled via `vueToast` rather than via this `toast` method, and, if possible, `vueToast` should probably be generally preferred for such needs.

app/javascript/groceries/components/Item.vue
```diff
@@ -30,11 +30,13 @@
...
     | &nbsp;
     span ({{item.needed}})
-  .delete.text-2xl.px-2.js-link.right.text-red-600.leading-unset(
+  .delete.text-2xl.js-link.right.text-red-600.leading-unset(
     v-if='ownStore'
-    @click="groceriesStore.destroyItem({ item })"
-    title='Delete item'
-  ) ×
+  )
+    button.px-2(
```
davidrunger, line 36: (We needed to make this into a proper button so that the feature spec could click on it (which also makes clear the accessibility need for this to be a proper button).)

app/javascript/groceries/components/DeletedItemToast.vue
```diff
@@ -0,0 +1,26 @@
...
+  },
+
+  restoreItemPath: {
+    type: String,
+    required: true,
+  },
+});
+
+function restoreItem() {
+  http.post(props.restoreItemPath);
```
davidrunger, line 24: I think I have previously made an effort to get all HTTP requests out of Vue components and into stores. Here, we are making an HTTP request in a Vue component. This feels better to me than the indirection of moving this into the store, just for the sake of following that principle. I think it's probably a good general principle, but I think this is a case where it doesn't make a lot of sense, particularly since there is no direct effect on / interaction with the data that is in the groceries Pinia store.

app/javascript/emoji_picker/components/CopiedEmojiToast.vue
```diff
@@ -0,0 +1,16 @@
+<template lang="pug">
+div
```
davidrunger, line 2: (Note that vue-toastification components must have a single root element; otherwise, there are errors/warning about being unable to pass a prop through or something.)

app/javascript/groceries/components/DeletedItemToast.vue
```diff
@@ -0,0 +1,26 @@
+<template lang="pug">
+div
```
davidrunger, line 2: (Note: as mentioned [here](https://github.com/davidrunger/david_runger/pull/4969/files#r1722724926), there must be only a single root component.)

lib/test/tasks/run_file_size_checks.rb
```diff
@@ -11,9 +11,10 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'check_ins*.js' => (183..193),
     'check_ins_index*.js' => (5..15),
     'copy_to_clipboard*.js' => (2..12),
-    'emoji_picker*.js' => (235..245),
-    'groceries*.css' => (93..103),
-    'groceries*.js' => (410..420),
+    'emoji_picker*.css' => (8..18),
+    'emoji_picker*.js' => (250..260),
+    'groceries*.css' => (108..118),
+    'groceries*.js' => (429..439),
```
davidrunger, line 17: It's a bit disappointing to be adding this page weight. To some extent, it's the result of the fact that we now have two toaster libraries, instead of just one. (We are adding `vue-toastification` in addition to `toastify-js`.) The advantage of `toastify-js` is that it can be used independently of Vue, which we sometimes need, but the advantage of `vue-toastification` is that it can render Vue components (in addition to simple strings), which is really helpful when trying to do something relatively complex in the toast, such as seen in the `DeletedItemToast.vue` component added in this PR. So, to some extent, I do think it makes sense to have both, for now. Also, I think that https://davidrunger.atlassian.net/browse/DEV-100 will make it so that, in these apps, we won't have double toaster libraries, anymore, so I think that these page weight might go back down when that ticket is implemented.

PR #4976, opened 2024-08-20
Include query counts (cached and uncached) in logs [DEV-101]

davidrunger:
This was added in Rails 7.2: https://github.com/rails/rails/pull/ 51457

(I checked, and it seems that there are no other notable values included in the available log data that we are also (up to now) failing to log.)

### DIFF
--- a/app/decorators/marriage_decorator.rb
+++ b/app/decorators/marriage_decorator.rb
@@ -5,6 +5,8 @@ class MarriageDecorator < Draper::Decorator
 
   memo_wise \
   def other_partner
-    partners.where.not(id: h.current_user).first
+    partner_ids = [partner_1_id, partner_2_id]
+    partner_id = (partner_ids - [h.current_user.id]).first
+    User.find(partner_id)
   end
 end
--- a/app/decorators/marriage_decorator.rb
+++ b/app/decorators/marriage_decorator.rb
@@ -7,6 +7,9 @@ class MarriageDecorator < Draper::Decorator
   def other_partner
     partner_ids = [partner_1_id, partner_2_id]
     partner_id = (partner_ids - [h.current_user.id]).first
-    User.find(partner_id)
+
+    if partner_id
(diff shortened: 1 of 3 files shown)


app/decorators/marriage_decorator.rb
```diff
@@ -5,6 +5,8 @@ class MarriageDecorator < Draper::Decorator
 
   memo_wise \
   def other_partner
-    partners.where.not(id: h.current_user).first
+    partner_ids = [partner_1_id, partner_2_id]
+    partner_id = (partner_ids - [h.current_user.id]).first
+    User.find(partner_id)
```
davidrunger: (Mostly unrelated change to try to optimize this query a bit (inspired by looking at our SQL queries more closely).)

lib/logs/log_formatter.rb
```diff
@@ -20,6 +20,8 @@ def initialize(data)
           when :format then 3
           when :params then 4
           when :status then 5
+          when :queries then 5.3
+          when :cached_queries then 5.6
```
davidrunger, line 24: I'm introducing these floats so as not to have to increment all of the other sort values below.

lib/logs/log_formatter.rb
```diff
@@ -85,6 +87,16 @@ def color_background_and_style(key, value)
...
+    when :queries
+      if value < 5
+        :green
+      elsif value < 12
+        :yellow
+      else
+        :red
+      end
+    when :cached_queries
+      :white
```
davidrunger, line 99: I'm not super sure whether cached queries are good or bad. On the one hand, it seems good, since we're at least not hitting the database. But, on the other hand, should we not even be sending these queries to ActiveRecord in the first place (and, instead, we should be memoizing something at the Ruby level that we currently aren't)? So, for now, I'm going to just cop out and color this "white" (actually a bit gray, in my terminal).

PR #4982, opened 2024-08-20
[emoji-picker] Treat space-separated query words as "AND" [EMO-6]

davidrunger:
# Before

image

# After

image

### DIFF
--- a/app/javascript/lib/composables/use_fuzzy_typeahead.ts
+++ b/app/javascript/lib/composables/use_fuzzy_typeahead.ts
@@ -27,7 +27,13 @@ export function useFuzzyTypeahead<T extends object>({
     if (query.value === '') {
       return searchables;
     } else {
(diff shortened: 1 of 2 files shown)


app/javascript/lib/composables/use_fuzzy_typeahead.ts
```diff
@@ -27,7 +27,8 @@ export function useFuzzyTypeahead<T extends object>({
     if (query.value === '') {
       return searchables;
     } else {
-      return map(fuse.value.search(query.value), 'item');
+      const searchResult = fuse.value.search(query.value);
```
davidrunger, line 30: Extracting this variable makes it easier to log it for debugging purposes, and I think that breaking it out and capturing it as a named variable also might help with readability, or at least probably doesn't hurt it.

PR #4984, opened 2024-08-21
[emoji-picker] Store boost preferences per user in database [EMO-9]

davidrunger:


app/javascript/lib/composables/use_fuzzy_typeahead.ts
```diff
@@ -8,19 +8,19 @@ export function useFuzzyTypeahead<T extends object>({
   maxMatches,
   fuseOptions,
 }: {
-  searchables: Array<T>;
+  searchables: Ref<Array<T>>;
```
davidrunger, line 11: By making this reactive, it's possible to see the boost effects on emoji searches immediately, rather than needing to refresh the page. There's actually even a live-typing effect, if one updates the boosted name while it is displayed in the search results.

app/javascript/lib/vue_toasts.ts
```diff
@@ -3,7 +3,7 @@ import { POSITION, useToast } from 'vue-toastification';
 import 'css/vue_toastification_with_overrides.scss';
 
 const DEFAULT_VUE_TOASTIFICATION_OPTIONS = Object.freeze({
-  position: POSITION.TOP_CENTER,
+  position: POSITION.TOP_RIGHT,
```
davidrunger, line 6: Top center is often kind of annoying and blocks content. Let's make the default top right, where I think it's less likely that there will be valuable content.

app/models/json_preference.rb
```diff
@@ -0,0 +1,32 @@
...
+class JsonPreference < ApplicationRecord
+  module Types
+    EMOJI_BOOSTS = 'emoji_boosts'.freeze
+  end
+
+  JsonPreference::Types.constants.each do |constant_name|
+    constant_value = JsonPreference::Types.const_get(constant_name)
+
+    scope(constant_value, -> { where(preference_type: constant_value) })
+  end
```
davidrunger, line 25: I feel bad about this metaprogramming, but I couldn't resist the DRYness.

app/models/user.rb
```diff
@@ -36,6 +36,17 @@ class User < ApplicationRecord
...
+
+  JsonPreference::Types.constants.each do |constant_name|
+    scope_name = JsonPreference::Types.const_get(constant_name).to_sym
+
+    has_one scope_name,
+      ->(user) { user.json_preferences.public_send(scope_name) },
+      class_name: 'JsonPreference',
+      inverse_of: :user,
+      dependent: :destroy
+  end
```
davidrunger, line 49: I feel bad about this metaprogramming, but I couldn't resist the DRYness.

davidrunger:
I guess that this `JsonPreference` model is fairly similar to the concept of this gem: https://github.com/ledermann/rails-settings .

PR #4996, opened 2024-08-22
Add and use typelizer [DEV-99]

davidrunger:
I think that this will make frontend development easier and faster, and it should also help to ensure that the contract/interface between the backend and frontend is respected and mutually understood.


app/javascript/groceries/store.ts
```diff
@@ -26,12 +27,6 @@ interface Nameable {
   name: string;
 }
 
-interface StoreAttributes {
-  name?: string;
-  notes?: string;
-  private?: boolean;
-}
```
davidrunger, line 33: (Nonessential, unrelated change. This is only used in one place; it seemed more appropriate to inline.)

app/javascript/groceries/store.ts
```diff
@@ -177,13 +172,11 @@ export const useGroceriesStore = defineStore('groceries', {
         }
       };
 
-      interface StoresResponse {
-        own_stores: Array<Store>;
-        spouse_stores: Array<Store>;
-      }
```
davidrunger, line 183: (Nonessential, unrelated change. This is only used in one place; it seemed more appropriate to inline.)

app/serializers/marriage_serializer.rb
davidrunger, line 1: I looked through the generated types to make sure that they are all being used. The Marriage type was not being used, which helped me to realize that this serializer is also not being used anymore.

app/serializers/quiz_serializer.rb
```diff
@@ -15,5 +15,8 @@
 #  index_quizzes_on_owner_id  (owner_id)
 #
 class QuizSerializer < ApplicationSerializer
-  attributes :hashid, :owner_id
+  attributes :owner_id
+
+  typelize 'string'
+  attribute(:hashid, &:hashid)
```
davidrunger, line 21: It's a tiny bummer that we need to break this out from `attributes` to type it, but really not that bad.

PR #5025, opened 2024-08-24
Use render_schema_json and JSON-schema-generated types more [DEV-99]

davidrunger:
... a lot more.

I tried to generate a JSON schema for pretty much every JSON response that isn't just an error(s). In at least a couple places, where the JSON response isn't actually being used by the frontend, I switched to `head :ok`, to make it more clear that the JSON isn't currently needed/used on the frontend.

I'm removing the `schema` keyword arg from the `render_schema_json` method. I think it makes significantly more sense to have one schema per action, even if there ends up being duplication between some of the actions. When the client side code makes a certain request, it makes a lot more sense for the type used to be specific for that request, rather than for the type name to randomly refer to some other request, just because that's how we set it up on the backend, I guess to reduce duplication.

Overall / stepping back, I'm not super sure whether this is worthwhile. I can imagine concluding that it's not, and undoing some of this. But, that being said, I can also imagine that this might possibly catch/prevent some bugs. I do feel like it fills what I have felt as a gap in the application's structure; it does feel satisfying to connect the backend and frontend in this way.

In development mode, the Api::BaseController tries to help out as much as it can, creating the necessary schema json file if it's missing, opening it in $EDITOR, and copying the JSON response data into the clipboard, for easy copy-pasting (after privacy sanitization) into https://jsonformatter.org/json-to-jsonschema .


app/controllers/api/base_controller.rb
```diff
@@ -1,26 +1,78 @@
...
-          strict: true,
-          validate_schema: true,
-        )
+        begin
+          JSON::Validator.fully_validate(
+            full_schema_path,
+            data.is_a?(String) ? data : data.to_json,
+            strict: true,
+            validate_schema: true,
+            clear_cache: true,
```
davidrunger, line 26: `clear_cache: true` is necessary to see changes in development without restarting the server.

app/controllers/api/base_controller.rb
```diff
@@ -1,26 +1,78 @@
 class Api::BaseController < ApplicationController
   class SchemaValidationError < StandardError ; end
 
-  def render_schema_json(data, schema: nil, **kwargs)
```
davidrunger, line 4: I'm removing the `schema` keyword argument. I think it's better for each action to have its own dedicated schema (particularly with respect to the names of the generated client-side types).

app/javascript/shared/helpers.ts
```diff
@@ -9,3 +9,10 @@ export function assert<T>(value: T | undefined | null): T {
 
   return value;
 }
+
+export function typesafeAssign<T extends object>(
```
davidrunger, line 13: TypeScript doesn't strictly enforce types for `Object.assign(...)`, i.e. it doesn't check that the objects are of the same (or similar) type. This function does.

app/javascript/shared/http.ts
```diff
@@ -22,18 +23,7 @@ const hooks: Hooks = {
...
-if (csrfToken) {
-  hooks['beforeRequest'] = [
-    (request) => {
-      request.headers.set('X-CSRF-Token', csrfToken);
-    },
-  ];
-}
-
-const kyApi = ky.extend({
+const kyApi = generalKyApi.extend({
```
davidrunger, line 26: I'm drying up our `ky` usage, so that we don't need to repeat the CSRF header configuration (as we were doing before) and so that I only need to specify `retry: 0` in one place in this PR.

app/javascript/types/index.ts
```diff
@@ -1,5 +1,9 @@
 export * from '@/types/serializers';
 
+export type Intersection<T, U> = {
+  [K in Extract<keyof T, keyof U>]: T[K];
+};
```
davidrunger, line 5: I use this type heavily in this PR; it gives a type that has only the common properties of its two arguments. (TypeScript has the `&` "intersection" operator, but I would call it a "merge" rather than an "intersection"; it combines the properties from both types, whether or not they are shared in common.)

app/javascript/types/responses/ItemCreateResponse.ts
```diff
@@ -0,0 +1,32 @@
+// This is a generated file. Do not edit this file directly.
```
davidrunger, line 1: The JSON schemas are pretty verbose, as it is, so it's an extra bummer that each one gets committed to the repo twice (once as a TypeScript file, as seen here, and once as a JSON file, which is copied into the corresponding TypeScript file). As mentioned [here](https://github.com/ThomasAribart/json-schema-to-ts/blob/e9ac5ee/documentation/FAQs/does-json-schema-to-ts-work-on-json-file-schemas.md), if this issue is ever addressed, then I guess we'll be able to remove this duplication: https://github.com/microsoft/TypeScript/issues/ 32063 .

lib/logs/log_formatter.rb
```diff
@@ -41,7 +41,7 @@ def call
         value = @data[key]
         key_value_string = format(key, value)
 
-        if Rails.env.local?
+        if Rails.env.development?
```
davidrunger: The code within this block causes a spec to fail, so we'll execute it now only in development.

spec/support/schemas/log_entries/create.json
```diff
@@ -0,0 +1,30 @@
...
+  "definitions": {
+    "Root": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "id": {
+          "type": "integer"
+        },
+        "data": {
+          "type": ["number", "string"]
```
davidrunger, line 13: Most of the content in these files was automatically generated by pasting example API responses into https://jsonformatter.org/json-to-jsonschema , but in some places (such as here, `"number"` => `["number", "string"]`) I had to manually modify some things, as surfaced by specs.

spec/support/schemas/stores/index.json
```diff
@@ -0,0 +1,94 @@
...
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/Store"
+          }
+        }
+      },
+      "required": ["own_stores", "spouse_stores"],
+      "title": "Root"
+    },
+    "Store": {
```
davidrunger, line 25: It's really cool and, I think, nice that https://jsonformatter.org/json-to-jsonschema is able to extract named type definitions (and, in turn, this type references another extracted `Item` type), rather than just building a single, big, deeply nested blob.

spec/support/schemas/stores/update.json
```diff
@@ -0,0 +1,45 @@
+{
+  "$schema": "http://json-schema.org/draft-06/schema#",
+  "$ref": "#/definitions/Root",
+  "definitions": {
+    "Root": {
+      "type": "object",
+      "additionalProperties": false,
```
davidrunger, line 7: Another thing I like about https://jsonformatter.org/json-to-jsonschema is that it specifies both `"additionalProperties": false` and lists all properties as `required`. Most/all other JSON-to-JSON-Schema generators I tried don't specify both of those.

app/javascript/types/index.ts
davidrunger: The format that I conventionally used is `Intersection<ModelTypeFromTypelizer, ResponseTypeFromJsonSchema>`.

PR #5031, opened 2024-08-25
[spec/quizzes] Add more page content expectations [QUIZ-4]

davidrunger:
It seems like maybe sometimes a response intended for the quiz owner gets processed in the session of the non-owner quiz participant? (If so, hopefully that's not something that can happen in real life, but rather is just a quirk of Capybara's session management.)

This change adds additional expectations about page content after taking actions in the browser. Hopefully this will slow things down a bit / keep things a bit more stable, so that responses are less likely to get mixed up as we move quickly back and forth between the two sessions.

I ran the test 60 times after making this change, and it flaked with the error that this change aims to fix (`TurboFrameMissingError`) zero times, whereas, when I did the same thing previously without this change, the spec flaked 5 times with `TurboFrameMissingError`. So, I think that's pretty encouraging.

Other flakiness remains in the spec being modified herein, but I think that I'll try to fix that in a separate PR.

### DIFF
--- a/app/models/quiz_question_answer.rb
+++ b/app/models/quiz_question_answer.rb
@@ -28,4 +28,5 @@ class QuizQuestionAnswer < ApplicationRecord
   has_many(:answering_participations, through: :selections, source: :participation)
 
   scope :correct, -> { where(is_correct: true) }
+  scope :incorrect, -> { where(is_correct: false) }
 end
--- a/spec/features/quizzes_spec.rb
+++ b/spec/features/quizzes_spec.rb
@@ -69,6 +69,7 @@ RSpec.describe 'Quizzes app' do
 
       # start the quiz
       click_on('Start quiz!')
+      expect(page).to have_text('Question 1')
 
       first_question, second_question = QuizQuestion.order(:created_at).last(2)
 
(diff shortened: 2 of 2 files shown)


app/models/quiz_question_answer.rb
```diff
@@ -28,4 +28,5 @@ class QuizQuestionAnswer < ApplicationRecord
   has_many(:answering_participations, through: :selections, source: :participation)
 
   scope :correct, -> { where(is_correct: true) }
+  scope :incorrect, -> { where(is_correct: false) }
```
davidrunger, line 31: I used to pretty much try to only add application code that is used by other application code. This `incorrect` scope is only used in a test, though. But I guess I am feeling a little more okay with that, now. I think that we can say that the purpose of application code is not solely to serve the end users. I think it's okay for at least some of it to be helpful to developers.

davidrunger:
Unfortunately, despite the early encouraging lack of failures, this doesn't seem to have actually fixed this flakiness (based on continuing to see these (and other) failures when running the specs locally).

That being said, these are probably good additions to the spec, anyway, so no harm done.

davidrunger:
Here's what actually fixed the flakiness that I had been seeing locally: https://github.com/davidrunger/david_runger/pull/5033 .

PR #5033, opened 2024-08-25
Add extra CSP sources in test when not in CI [QUIZ-4]

davidrunger:
Otherwise, when running tests locally while using the Vite development server, we sometimes get CSP errors, and these can cause problems (such as flaky specs).

Also, suppress Vite "connected" message, since we are now able to connect successfully, since the CSP is no longer a blocker.

Also, add a one second `sleep` to a quizzes feature spec. Without this, the login takes about 10 seconds, for some reason. Also, I think that this `sleep` makes these spec much more reliable when run on my machine (though it already seem reliable in CI).

### DIFF
--- a/spec/features/quizzes_spec.rb
+++ b/spec/features/quizzes_spec.rb
@@ -33,6 +33,10 @@ RSpec.describe 'Quizzes app' do
       # quiz show page expectations without participant
       expect(page).not_to have_text(participant_name)
 
+      # NOTE: Sleeping for 1 second here makes the login below about 10 seconds
+      # faster, and it might also help these specs not to flake on my machine.
+      sleep(1)
+
       # a participant joins
       using_session('Quiz participant session') do
         wait(20.seconds).for do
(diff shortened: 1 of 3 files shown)


spec/features/quizzes_spec.rb
```diff
@@ -33,6 +33,10 @@
       # quiz show page expectations without participant
       expect(page).not_to have_text(participant_name)
 
+      # NOTE: Sleeping for 1 second here makes the login below about 10 seconds
+      # faster, and it might also help these specs not to flake on my machine.
+      sleep(1)
```
davidrunger, line 38: Below, within `using_session('Quiz participant session')`, we call `sign_in(participant)`.

Without this sleep, I think that call maybe sometimes (initially?) happens within the context of this main Capybara session. I think that maybe the big advantage of this sleep is that it makes sure that this main session is "done" before we enter into the `using_session` block and sign in the quiz participant. Therefore, without this sleep, we sometimes get the quiz owner and quiz participant sessions mixed together, which can produce some really confusing failures.

PR #5045, opened 2024-08-27
Check bootstrap data against JSON schemas [DEV-102]

davidrunger:
This extends the work started in #5023 and #5025 now to also validate the JSON schema of bootstrap data and use the JSON schemas as TypeScript type definitions.

As with those PRs, I am not totally sure that this is worth it. Let's see how it goes.


app/controllers/api/base_controller.rb
```diff
@@ -1,78 +1,5 @@
 class Api::BaseController < ApplicationController
-  class SchemaValidationError < StandardError ; end
```
davidrunger, line 2: This code was basically moved to `SchemaValidatable` (a controller concern) and `JsonSchemaValidator` (a PORO).

app/controllers/concerns/bootstrappable.rb
```diff
@@ -8,13 +8,18 @@ module Bootstrappable
...
 
   def bootstrap(data = {})
-    # Adding a `nonce` ensures that the bootstrap data is always unique, which is useful because
-    # that ensures that turbo will always evaluate the `<script>` tag(s) within which it's wrapped.
-    @bootstrap_data ||= {
-      nonce: SecureRandom.uuid,
-      current_user: current_user && UserSerializer::Basic.new(current_user),
-    }
-    @bootstrap_data.merge!(data)
+    if data.present?
```
davidrunger, line 11: We used to always render bootstrap data, even when it wasn't needed/used. With this change, we can now check whether `@bootstrap_data` is present, and, if not, skip rendering it altogether. That way, when the bootstrap data is not used (and now won't be rendered into the page), we won't validate it and don't need to add a schema for it.

app/javascript/check_ins/store.ts
```diff
@@ -9,9 +10,11 @@ import { Bootstrap, NeedSatisfactionRating } from './types';
 
 declare const Routes: typeof RoutesType;
 
+const bootstrap = untypedBootstrap as Bootstrap;
+
 export const useCheckInsStore = defineStore('check-ins', {
   state: () => ({
-    ...(window.davidrunger.bootstrap as Bootstrap),
```
davidrunger, line 14: For the sake of DRY and also (at least in some cases) better type safety, I removed all remaining direct references to `window.davidrunger.bootstrap`. Now, we only access bootstrap data via `@/lib/bootstrap`.

app/javascript/groceries/Groceries.vue
```diff
@@ -11,13 +11,13 @@
 <script setup lang="ts">
 import { Connection } from '@rails/actioncable';
 import Cookies from 'js-cookie';
-import { get } from 'lodash-es';
```
davidrunger, line 14: `get` is not type-safe (it's return value is `any`), so removing its usage seems like a win.

PR #5053, opened 2024-08-29
[log_formatter] Don't reference Rainbow in production

davidrunger:
It's not available there.

We are seeing this error in papertrail logs:

> Could not log "process_action.action_controller" event. NameError: uninitialized constant Logs::LogFormatter::Rainbow

### DIFF
--- a/lib/logs/log_formatter.rb
+++ b/lib/logs/log_formatter.rb
@@ -36,6 +36,7 @@ class Logs::LogFormatter < Lograge::Formatters::KeyValue
   # rubocop:enable Metrics/CyclomaticComplexity
 
   # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
   def call
     fields_to_display(@data).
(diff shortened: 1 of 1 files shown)


lib/logs/log_formatter.rb
```diff
@@ -36,13 +36,14 @@ def initialize(data)
...
 
   # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/PerceivedComplexity
   def call
     fields_to_display(@data).
       map do |key|
         value = @data[key]
         key_value_string = format(key, value)
 
-        if Rails.env.development? || !Rainbow.enabled
```
davidrunger, line 45: This line was changed to this here: https://github.com/davidrunger/david_runger/pull/5025/files#diff-19d5e58264903a198201b6bf0e4b0a9ad55f0f0b7323a238ada7f3e9881d9661R45

Could this simply be `Rails.env.development?`? Maybe. I'm not sure. I'll just go with this, for now, though.

PR #5061, opened 2024-08-30
[ErrorSubscriber] Handle uncaught errors, as well

davidrunger:
Also, bump Rollbar from 3.5.2 to 3.6.0. Version 3.6.0 includes support for using Rollbar via a Rails error subscriber class. We ultimately aren't using this functionality, though, and instead are still rolling our own error subscriber, so that we can also write the errors to the logs. The key thing that I took away from the Rollbar PR implementing that functionality ( https://github.com/rollbar/rollbar-gem/pull/ 1161 ), though, is its use of the `capture_uncaught` Rollbar configuration option, which I am using in this PR (and which I think I wasn't previously aware of), so that we can now have the Rails error subscriber handle both caught and uncaught exceptions, without Rollbar also sending any uncaught exceptions.

Also, use `git` to determine a SHA for Rollbar as a fallback if `GIT_REV` is not present. This makes it easier to test in development without setting `GIT_REV`, and is probably not a bad idea to have available as a fallback in other contexts, too.

### DIFF
--- a/config/initializers/rollbar.rb
+++ b/config/initializers/rollbar.rb
@@ -1,5 +1,5 @@
 Rollbar.configure do |config|
-  code_version = ENV.fetch('GIT_REV', nil)
+  code_version = ENV.fetch('GIT_REV', `git log --format=format:%H | head -n 1`.rstrip)
 
   access_token =
     case Rails.env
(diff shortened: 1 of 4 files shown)


config/initializers/rollbar.rb
```diff
@@ -1,5 +1,5 @@
 Rollbar.configure do |config|
-  code_version = ENV.fetch('GIT_REV', nil)
+  code_version = ENV.fetch('GIT_REV', `git log --format=format:%H | head -n 1`.rstrip)
```
davidrunger, line 2: See https://github.com/davidrunger/david_runger/pull/5063 for an improvement on this.

PR #5085, opened 2024-09-03
Use delegated types for log entries [LOG-21]

davidrunger:
Using delegated types might add some net significant lines of code (though it's hard to tell for sure from the diff, since a fair amount of the diff is incidental additional and deletions), and using delegated types is arguably, in at least some sense, somewhat more complex. In particular, there is some delicacy and additional complication when creating or updating records, because now we must update two separate records, rather than updating just a single record, as we did before.

However, that being said, I do kind of think it's an improvement / the right way to go. Our old data structure was decently confusing, in its own right, and I always had the nagging feeling that it is somehow "wrong". This new approach feels a little more "right" to me. For one thing, the data structure is DRYer; for example, after cleanup (though not as of this PR) we won't duplicate the `note` column across multiple tables.


Gemfile.lock
```diff
@@ -565,7 +565,7 @@ GEM
     runger_config (5.2.0)
       activesupport (>= 7.1.2)
     runger_email_reply_trimmer (0.3.0)
-    runger_style (2.16.0)
+    runger_style (2.17.0)
```
davidrunger, line 568: This tells RuboCop to ignore some of the otherwise too-long lines in `annotate` comments about an index.

app/models/log.rb
```diff
@@ -44,17 +43,8 @@ class Log < ApplicationRecord
...
-    dependent: :destroy,
-    inverse_of: :log
-
-  has_many :text_log_entries,
-    class_name: 'LogEntries::TextLogEntry',
-    dependent: :destroy,
-    inverse_of: :log
-
-  has_many :log_shares, dependent: :destroy
+  has_many :log_entries, dependent: :destroy, inverse_of: :log
```
davidrunger, line 46: I think that explicitly specifying this `inverse_of` option is essential for something to work correctly.

app/models/log.rb
```diff
@@ -64,36 +54,24 @@ class Log < ApplicationRecord
...
           (
-            MAX(number_log_entries.created_at) IS NOT NULL
-            AND (logs.reminder_last_sent_at IS NULL OR logs.reminder_last_sent_at < MAX(number_log_entries.created_at))
-            AND EXTRACT(EPOCH FROM (NOW() - MAX(number_log_entries.created_at))) >= logs.reminder_time_in_seconds
-          )
-          OR
-          (
-            MAX(text_log_entries.created_at) IS NOT NULL
-            AND (logs.reminder_last_sent_at IS NULL OR logs.reminder_last_sent_at < MAX(text_log_entries.created_at))
-            AND EXTRACT(EPOCH FROM (NOW() - MAX(text_log_entries.created_at))) >= logs.reminder_time_in_seconds
```
davidrunger, line 88: The simplification here is one of the upsides of this approach.

app/models/log_entry.rb
```diff
@@ -2,26 +2,32 @@
...
 class LogEntry < ApplicationRecord
   include JsonBroadcastable
 
-  self.abstract_class = true
-
   belongs_to :log
   has_one :user, through: :log
+  delegated_type :log_entry_datum,
+    types: %w[NumberLogEntryDatum TextLogEntryDatum],
+    inverse_of: :log_entry
```
davidrunger, line 25: I think this `inverse_of` might be essential for something to work correctly.

app/models/log_entry.rb
```diff
@@ -2,26 +2,32 @@
...
+
+  validates :log_entry_datum, presence: true
+  validates_associated :log_entry_datum
```
davidrunger, line 28: I don't love the behind-the-scenes/implicit "magic" (and inescapability) of `validates_associated`, but I think it's worth it, here.

db/datamigrate/create_log_entries_from_data.rb
```diff
@@ -0,0 +1,27 @@
+class CreateLogEntriesFromData
```
davidrunger, line 1: We'll need to execute this datamigration ASAP after deploying; logs will not show data in the web UI until we do.

lib/tasks/auto_annotate_models.rake
```diff
@@ -1,35 +0,0 @@
-# NOTE: only doing this in development as some production environments (Heroku)
```
davidrunger, line 1: The deletion of this file is mostly unrelated. Basically, I think it wasn't doing anything (for our workflows).

spec/factories/log_entries.rb
```diff
@@ -2,26 +2,22 @@
 #
 # Table name: log_entries
```
davidrunger, line 3: This table didn't even exist before. At least, not recently. I think it used to exist a long time ago, which was I think how annotate was able to initially create this comment.

PR #5134, opened 2024-09-06
Use ransack fork to avoid error calling ActiveRecord::Base.connection

davidrunger:
fixes rb#673

https://app.rollbar.com/a/davidjrunger/fix/item/davidrunger/673


davidrunger:
See https://github.com/davidrunger/ransack/pull/1 .

PR #5171, opened 2024-09-13
Allow permanent DB connection for database_consistency and immigrant

davidrunger:
I am broadening the monkeypatch scope from just in test to also include development, because I want to be able to run these checks in both development and test without error.

Also, change the containing module from `Runger` (provided (but shouldn't be?) by the `runger_config` gem?) to `DavidRunger`, which is defined in `config/application.rb`.

### DIFF
--- a/config/initializers/monkeypatches/active_record_connection_handling.rb
+++ b/config/initializers/monkeypatches/active_record_connection_handling.rb
@@ -1,7 +1,7 @@
-if Rails.env.test?
+if Rails.env.local?
   # Code being patched:
   # https://github.com/rails/rails/blob/v7.2.1/activerecord/lib/active_record/connection_handling.rb#L259-L281
-  module Runger::ConnectionHandlingMonkeypatch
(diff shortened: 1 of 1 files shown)


config/initializers/monkeypatches/active_record_connection_handling.rb
```diff
@@ -15,26 +15,28 @@ def connection
     end
 
     def check_calling_connection_allowed
-      case ActiveRecord.permanent_connection_checkout
-      when :deprecated
-        # :nocov:
-        ActiveRecord.deprecator.warn(<<~MESSAGE)
-          Called deprecated `ActiveRecord::Base.connection` method.
+      unless $PROGRAM_NAME.match?(/\bdatabase_consistency\b/) || caller.any?(/\bimmigrant\b/)
```
davidrunger, line 18: The better thing to do would be to submit PRs to these two gems (`database_consistency` and `immigrant`) to have them stop using `ActiveRecord::Base.connection`, but I don't feel like doing that.

PR #5208, opened 2024-09-20
Add MIT license

davidrunger:
Like Rails ( https://github.com/rails/rails/pull/ 47467 ), I am omitting the year range that is suggested by https://choosealicense.com/licenses/mit/ (which is where I otherwise copied this license/template from).


davidrunger:
It just seems like the right thing to do, from an open source perspective.

PR #5434, opened 2024-11-04
Store JWT secret in ENV var (not Rails credentials)

davidrunger:
This is somewhat of a follow-up to ef138c27 / #4731 .

Earlier this year, I lost the encryption secret for the development credentials file. Rather than rebuild a new credentials file (which sort of demonstrates how credentials can be a bit of a pain), instead, we'll continue herein migrating from Rails credentials to ENV vars.

I have just set a `JWT_SECRET` env var in production (to the same value as `Rails.application.credentials.jwt_secret` was), so this should be a smooth, no-downtime deployment.


spec/controllers/proposals_controller_spec.rb
```diff
@@ -6,20 +6,19 @@
...
 
-    context 'when a jwt_secret credential is available' do
-      before do
-        expect(Rails.application.credentials).
-          to receive(:jwt_secret!).
-          at_least(:once).
-          and_return('xyzdef2481632')
+    context 'when JWT_SECRET is set' do
+      around do |spec|
+        ClimateControl.modify(JWT_SECRET: 'zyx432') do
```
davidrunger, line 19: Shortening the secret (vs `xyzdef2481632` before) makes it so that gitleaks doesn't pick this up as a "real" secret and warn about it.

PR #5471, opened 2024-11-08
Reallow ApplicationRecord::Base.connection

davidrunger:
https://github.com/activeadmin/activeadmin/pull/ 8474 is not going to be merged (at least, not in the near future), and so this won't be fixed upstream in activeadmin anytime soon. Therefore, let's give up and get back onto upstream activeadmin (and ransack) now by reallowing `connection`.


davidrunger:
Another motivation for this change is that I am thinking about switching from `annotate` to `annotaterb`, since the former is not compatible with Rails 8.0 (https://github.com/ctran/annotate_models/issues/ 1032). The latter seems to call `ApplicationRecord::Base.connection`, though. So, making this change might help us to switch to `annotaterb` (and thus to Rails 8.0).

PR #5508, opened 2024-11-14
Remove sprockets-rails

davidrunger:
Do we need this anymore? I guess that Rails 8.0 transitions from sprockets to propshaft? We still have sprockets. Maybe we don't need it? Or, if we do, maybe it should be migrated to propshaft?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,7 +46,6 @@ gem 'runger_actions'
 gem 'runger_email_reply_trimmer'
 gem 'sassc' # used by ActiveAdmin asset pipeline
 gem 'sidekiq'
-gem 'sprockets-rails'
 gem 'strip_attributes'
 gem 'typelizer', github: 'davidrunger/typelizer'
 gem 'vite_rails'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -619,13 +619,6 @@ GEM
     spring-watcher-listen (2.1.0)
       listen (>= 2.7, < 4.0)
       spring (>= 4)
-    sprockets (4.2.1)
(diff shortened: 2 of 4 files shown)


davidrunger:
I don't think that this is going to work, because I think that activeadmin seems to break with this removal of sprockets.

```
~/code/david_runger  remove-sprockets ✔ 4↥  12:16:00 PM
❯ rs
=> Booting Puma
=> Rails 8.0.0 application starting in development 
=> Run `bin/rails server --help` for more startup options
Exiting
/home/david/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/railties-8.0.0/lib/rails/railtie/configuration.rb:109:in `method_missing': undefined method `assets' for an instance of Rails::Application::Configuration (NoMethodError)
Did you mean?  asset_host
	from /home/david/.rbenv/versions/3.3.6/lib/ruby/gems/3.3.0/gems/activeadmin-3.2.5/lib/active_admin/engine.rb:13:in `block (2 levels) in <class:Engine>'
```

A similar error also occurred in CI ([link](https://github.com/davidrunger/david_runger/actions/runs/11843099251/job/33003267084?pr=5508#step:11:524)).

I think I am going to try opening a new PR migrating from sprockets to propshaft. Maybe that will work.

davidrunger:
> I think I am going to try opening a new PR migrating from sprockets to propshaft. Maybe that will work.

Nevermind. The [upgrade process](https://github.com/rails/propshaft/blob/main/UPGRADING.md) is a lot more than I want to take on.

PR #5612, opened 2024-12-04
Update gems

davidrunger:
`bundle update`


davidrunger:
This is going to fail because of rubocop, as mentioned here: https://github.com/davidrunger/david_runger/pull/5314#issuecomment-2416973954 . Closing.

PR #5781, opened 2025-01-12
Serve '/.well-known/traffic-advice' to permit all prefetching proxies

davidrunger:
https://developer.chrome.com/blog/private-prefetch-proxy/#traffic


davidrunger:
See also the comment that I added here: https://webmasters.stackexchange.com/questions/138033/what-is-well-known-traffic-advice-directory#comment204820_139570

PR #5791, opened 2025-01-13
Send original logging container ID as host/system to Papertrail

davidrunger:
Previously, the container id of the vector container was being sent as the host/system, which is not particularly helpful.


davidrunger:
One possible downside of this is that it will greatly fragment the "systems" in Papertrail (over time as we deploy changes and also since each process/program will also be its own "system" now). It might be better to just always delete the `host`, in which case I think Papertrail's default behavior is to log the originating IP address as the "system". I think that I will open a new PR to do that.

PR #5792, opened 2025-01-13
[vector] Don't send host/system to Papertrail

davidrunger:
Previously, the container id of the vector container was being sent as the host/system, which is not particularly helpful, and which I think might have some downsides. When no host/system is provided, I think Papertrail will just log the sending IP address as the system, which I think actually works pretty well, so this change should bring up back to that.

### DIFF
--- a/vector/vector.yaml
+++ b/vector/vector.yaml
@@ -13,11 +13,13 @@ transforms:
       match(string!(.container_name), r'-(clock|initialize_database|web|worker)-\d+')
 
   # Add brief service name to log data
-  papertrail_services_with_brief_service_name:
+  papertrail_services_remapped:
     type: remap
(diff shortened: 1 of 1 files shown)


vector/vector.yaml
```diff
@@ -13,11 +13,13 @@ transforms:
       match(string!(.container_name), r'-(clock|initialize_database|web|worker)-\d+')
 
   # Add brief service name to log data
-  papertrail_services_with_brief_service_name:
+  papertrail_services_remapped:
     type: remap
     inputs:
       - papertrail_services
     source: |
+      del(.host)
```
davidrunger, line 21: We were sending this as `host`, but I think it is called "System" in the Papertrail UI.

PR #5793, opened 2025-01-13
[docker-compose] Add '--watch-config' to vector command

davidrunger:
This way, the service will reload itself if/when the configuration file is changed. (This doesn't happen automatically, because the file is not part of a Dockerfile definition, but rather is provided to the service via a Docker volume.)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -175,7 +175,11 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock:ro
       - ./vector/vector.yaml:/etc/vector/vector.yaml:ro
     command:
-      ['--config', '/etc/vector/vector.yaml', '--require-healthy', 'true']
+      - '--config'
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -175,7 +175,11 @@ services:
       - /var/run/docker.sock:/var/run/docker.sock:ro
       - ./vector/vector.yaml:/etc/vector/vector.yaml:ro
     command:
-      ['--config', '/etc/vector/vector.yaml', '--require-healthy', 'true']
+      - '--config'
+      - '/etc/vector/vector.yaml'
+      - '--require-healthy'
+      - 'true'
+      - '--watch-config'
```
davidrunger, line 182: (I'm also breaking out this array into a vertical YAML format, just because the arguments list is getting kind of long and this will make it somewhat easier to add other arguments in the future, if needed.)

PR #5796, opened 2025-01-14
Remove RAILS_MAX_THREADS fallback and change its value from 3 to 2

davidrunger:
Motivation for removing the fallback: It essentially DRYs things up. Previously, we had the same fallback value in three or four different files and would need to update it in each of those places whenever we wanted to change it.

Motivation for reducing from 3 to 2: Hopefully this will save us a little bit of memory, which I think that we might want/need, partially because I am thinking about adding yet more services to our Docker Compose setup (specifically Loki and Grafana), in addition to having recently added Vector.

Sidekiq will use RAILS_MAX_THREADS as its default concurrency ("Starting in Rails 5, `RAILS_MAX_THREADS` can be used to configure Rails and Sidekiq concurrency", [source][1]), so I'm also removing the concurrency specifications from `config/sidekiq.yml`, since they are basically WET with the specifications of RAILS_MAX_THREADS in `.env.production` (added in this change) and `.env.development`.

[1]: https://github.com/sidekiq/sidekiq/wiki/Advanced-Options#concurrency


davidrunger:
Memory before on the server:
- 654M total (`used` in `htop`)
- 151 MiB worker (per `docker stats`)
- 170 MiB web (per `docker stats`)

We'll see how this changes post-deploy.

Note, however, that I have sometimes seen significantly lower numbers for the worker and web memory usage, so I'm not sure the above figures are representative.

---

**Update.** Speaking of not being representation, here's another data point that I took just now:
- worker: 105 MiB
- web: 145 MiB

---

Another update:
- worker: 108
- web: 131

---

- worker: 57 MiB
- web: 91 MiB

This was shortly after a failed deploy attempt. I think that the deploy pulls away some memory from the running Docker containers.

PR #5798, opened 2025-01-14
[Dockerfile] Add 'libyaml-dev' to packages installed for build step

davidrunger:
After running `docker system prune --force --all --volumes && docker volume prune --all --force`, I was getting an error when trying to build the Docker image for the project, specifically during the `bundle install` command.

I'm not sure why this issue has emerged now. Maybe it has to do with bumping Ruby to 3.4.1 relatively recently? Maybe that image doesn't include libyaml-dev by default, but previous Ruby images did?

Anyway, adding this seems to fix the issue (although there is at least one unresolved issue in a later build step, which I will try to fix in a subsequent PR).


davidrunger:
> there is at least one unresolved issue in a later build step, which I will try to fix in a subsequent PR

Here it is: #5797 (opened before this PR, but not yet merged)

PR #5814, opened 2025-01-15
[vector] Only send logs to Papertrail if HOST/PORT env var is present

davidrunger:
This will make it more straightforward to run Docker containers locally without sending logs to Papertrail (e.g. by simply commenting out PAPERTRAIL_HOST_AND_PORT in `.env.vector.local`).

### DIFF
--- a/vector/vector.yaml
+++ b/vector/vector.yaml
@@ -30,7 +30,7 @@ sinks:
     encoding:
       codec: text
     endpoint: '${PAPERTRAIL_HOST_AND_PORT}'
-    if: '${PAPERTRAIL_HOST_AND_PORT:-false}' # Defaults to false if env var not set.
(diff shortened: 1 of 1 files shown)


vector/vector.yaml
```diff
@@ -27,10 +27,11 @@ transforms:
 sinks:
   # Send logs to Papertrail
   papertrail:
-    type: papertrail
-    inputs:
-      - papertrail_services_remapped
-    endpoint: '${PAPERTRAIL_HOST_AND_PORT}'
     encoding:
       codec: text
+    endpoint: '${PAPERTRAIL_HOST_AND_PORT}'
```
davidrunger, line 34: (I also alphabetized the other configuration entries. The only substantive change is adding the `if` below.)

PR #5821, opened 2025-01-16
[development docker] Don't send DataMonitorMailer.expectation_not_met

davidrunger:
We cannot send emails in development mode when running in Docker, because in development the email service is letter_opener, which is a development-only gem, whereas Docker installs only production gems. (I guess that letter_opener cannot be made to work within Docker, anyway, since the `worker` service machine does not have a browser?)

(Due to using a seeded database when running Docker in development, many of the data monitor checks will fail, so we would be trying to send these emails a decent amount of the time. For example, we have only one user, rather than the 4..100 expected.)

### DIFF
--- a/app/workers/data_monitors/base.rb
+++ b/app/workers/data_monitors/base.rb
@@ -1,5 +1,6 @@
 class DataMonitors::Base
   include Throttleable
+  prepend MemoWise
 
   TIME_BETWEEN_EMAIL_ALERTS = 1.day.freeze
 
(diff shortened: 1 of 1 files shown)


davidrunger:
Actually, it occurs to me that this will be an issue with any emails that we attempt to send in development when running in Docker. I think I will try to find a more generic solution, probably/hopefully a mailer `delivery_method` that will just log some info about the email to stdout (and which I will use when `Rails.env.development? && IS_DOCKER` is true).

PR #5825, opened 2025-01-16
[nginx/davidrunger.com.conf] Add/tweak comments and indentation

davidrunger:
These changes should be totally non-substantive.

### DIFF
--- a/config/nginx/sites-enabled/davidrunger.com.conf
+++ b/config/nginx/sites-enabled/davidrunger.com.conf
@@ -16,7 +16,7 @@ server {
     access_log              /var/log/nginx/access.log combined buffer=4k flush=1s;
     error_log               /var/log/nginx/error.log warn;
 
-    # SSL certificate renewals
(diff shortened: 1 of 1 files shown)


config/nginx/sites-enabled/davidrunger.com.conf
```diff
@@ -48,6 +48,8 @@ server {
     # logging
     access_log  /var/log/nginx/access.log combined buffer=4k flush=1s;
     error_log   /var/log/nginx/error.log warn;
+
+    # Ownership verification when requesting SSL certificates
```
davidrunger, line 52: I'm not 100% sure that this even gets used (here in the port 80 / HTTP `server` block). I know that SSL renewal requests are made with HTTPS. I am not sure whether requests for fresh certificates are made with HTTPS or HTTP. If those also are made with HTTPS, then this might not be needed. But performing that analysis is outside of scope here. Yet I wanted to add an explanatory comment.

PR #5827, opened 2025-01-16
Store logs in Loki and view them via Grafana (via NGINX) [DEV-120]

davidrunger:
Docker's built-in json-file logging seems to not work very well, at least when also using log rotation. For example, `docker compose logs <service_name>` only shows logs for the current container, even if that container was just booted up recently. There are other problems. And we can't currently afford to send all logs (including NGINX) to Papertrail. So, we don't currently have a good solution that covers NGINX. Loki/Grafana will provide that.

A side benefit / win-win is that it would also be nice to have Grafana available for viewing metrics (aggregated/stored via StatsD, Graphite, and/or whatever), and this will lay some of the groundwork for that by setting up Grafana.

### DIFF
--- a/bin/server/bootstrap-letsencrypt.sh
+++ b/bin/server/bootstrap-letsencrypt.sh
@@ -5,7 +5,7 @@
 
 # shellcheck disable=SC2128
 
-domains=(davidrunger.com www.davidrunger.com)
+domains=(davidrunger.com grafana.davidrunger.com www.davidrunger.com)
 rsa_key_size=4096
 data_path="./ssl-data/certbot"
 email="davidjrunger@gmail.com" # Adding a valid address is strongly recommended
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -65,6 +65,18 @@ services:
       interval: 5m
       timeout: 1s
       retries: 1
+  grafana:
+    depends_on:
+      - loki
+    env_file:
+      - .env.grafana.local
+    expose:
+      - '3000'
+    image: grafana/grafana:latest
+    networks:
+      - internal
+    volumes:
+      - grafana-data:/var/lib/grafana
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -177,6 +197,9 @@ services:
     env_file:
       - .env.vector.local
     image: timberio/vector:latest-alpine
+    networks:
+      - external
+      - internal
```
davidrunger, line 202: (`external` for sending to Papertrail, `internal` for sending to `loki`)

vector/vector.yaml
```diff
@@ -4,6 +4,15 @@ sources:
     type: docker_logs
 
 transforms:
+  docker_with_grafana_nginx_separated:
+    type: remap
+    inputs:
+      - docker
+    source: |
+      if .label."com.docker.compose.service" == "nginx" && match(string!(.message), r'^\[grafana\]') {
+        .label."com.docker.compose.service" = "nginx-grafana"
```
davidrunger, line 13: I feel a little guilty about overwriting the service with a name that's not really one of our service names, but I think it's worth it to keep the main app vs Grafana NGINX logs separate.

bin/server/bootstrap-letsencrypt.sh
```diff
@@ -5,7 +5,7 @@
 
 # shellcheck disable=SC2128
 
-domains=(davidrunger.com www.davidrunger.com)
+domains=(davidrunger.com grafana.davidrunger.com www.davidrunger.com)
```
davidrunger, line 8: (In Cloudflare, I just edited the `ww2` A record entry (which I had to work around a LetsEncrypt rate limit that I was bumping up against the last time that I tried to renew SSL certificates) to `grafana`.)

PR #5838, opened 2025-01-16
[roll-out-web] Log results of all curl requests to new web container

davidrunger:
Lately, we have been getting deployment failures that seem to be occurring on the line of code being modified herein. Example: https://github.com/davidrunger/david_runger/actions/runs/12812774650/job/35725480230

To give better visibility into what might be causing those deployment failures, this change adds more verbose logging that shows the result of each `curl` request to the new web container.

Additionally, we herein increase the delay between `curl` requsets from 1 second to 2 second, thereby increasing the overall allowed time from 30 seconds to 60 seconds. Motivation: it's possible that the cause of the deployment failures is simply that the new container is not becoming available quickly enough (i.e. within 30 seconds). If that is the case, then increasing the allowed time should be a straightforward solution. If that is not the case (which might be revealed or at least hinted at by this new logging), then we can go back to 30 seconds or even less.

### DIFF
--- a/bin/server/roll-out-web.sh
+++ b/bin/server/roll-out-web.sh
@@ -32,7 +32,47 @@ scale_web 2
 new_container_id=$(docker ps --filter name=web --quiet | head -1)
 new_container_ip=$(docker inspect --format '{{.NetworkSettings.Networks.david_runger_internal.IPAddress}}' "$new_container_id")
 echo_iso8601 "Curling new container IP until success."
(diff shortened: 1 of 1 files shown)


davidrunger:
Yes, the [deploy of this PR](https://github.com/davidrunger/david_runger/actions/runs/12814569640/job/35731430688) (which failed for a (presumably unrelated) `certbot`-related reason) took 13 attempts (26 seconds), so I am guessing that the main cause of deployment failures that we were seeing was that they were timing out due to taking more than 30 seconds. So, this change will resolve the issue, I think.

PR #5854, opened 2025-01-18
[docker-compose/web] Increase start_period from 30s to 60s

davidrunger:
We [wait up to 60 seconds][1] for the service to become healthy, so we should probably make the start period at least that long. Otherwise, I think that one we go through the first 30 seconds of checks then the remaining the 30 seconds are essentially wasted, if `web` wasn't healthy after those first 30 seconds, because it won't become healthy within that next 30 seconds.

[1]: https://github.com/davidrunger/david_runger/blob/8e1590df/bin/server/verify-expected-services.sh/#L7-L8

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -220,7 +220,7 @@ services:
       - '3000'
     healthcheck:
       test: curl --silent --output /dev/null --fail localhost:3000/up
-      start_period: 30s
+      start_period: 60s
       start_interval: 1s
       interval: 1m
       timeout: 1s


davidrunger:
> I think that on[c]e we go through the first 30 seconds of checks then the remaining the 30 seconds are essentially wasted, if web wasn't healthy after those first 30 seconds, because it won't become healthy within that next 30 seconds.

I think that this might explain deploy failures like this one (where `web` was still not considered healthy after 30 retries with 2 seconds between each): https://github.com/davidrunger/david_runger/actions/runs/12835217146/job/35794332318

PR #5857, opened 2025-01-19
Add Prometheus and node-exporter to track system metrics [DEV-121]

davidrunger:
This was astonishingly easy and straightforward. When combined with [this dashboard][1], this provides pretty thorough visibility into pretty much any system metric that one might care about.

[1]: https://grafana.com/grafana/dashboards/1860-node-exporter-full/

### DIFF
--- a/prometheus.yml
+++ b/prometheus.yml
@@ -1,0 +1,10 @@
+global:
+  scrape_interval: 20s
+  evaluation_interval: 20s
+
+scrape_configs:
+  # NOTE: This name is required for this dashboard:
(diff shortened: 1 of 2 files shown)


prometheus.yml
```diff
@@ -0,0 +1,10 @@
+global:
+  scrape_interval: 20s
+  evaluation_interval: 20s
```
davidrunger, line 3: I think that the default is 1 minute ([source](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration-file)). I am interested in trying to get somewhat better resolution data than that.

PR #5863, opened 2025-01-19
[docker-compose/nginx] Provide host for healthcheck

davidrunger:
Without this, the request will be rejected by NGINX due to [this][1] and [this][2].

[1]: https://github.com/davidrunger/david_runger/blob/c7e93885/config/nginx/nginx.conf/#L76-L79
[2]: https://github.com/davidrunger/david_runger/blob/c7e93885/config/nginx/nginxconfig.io/enforce-host.conf/#L1-L2

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -126,7 +126,8 @@ services:
     env_file:
       - .env.nginx.local
     healthcheck:
-      test: curl -s -o /dev/null -w "%{http_code}" "http://localhost/nginx-health" | grep -q "^200$" || false
+      test: |-
+        curl -s -o /dev/null -w "%{http_code}" --header 'Host: davidrunger.com' http://localhost/nginx-health | grep -q "^200$" || false
       start_period: 60s
       start_interval: 2s
       interval: 10s


docker-compose.yml
```diff
@@ -126,7 +126,8 @@ services:
     env_file:
       - .env.nginx.local
     healthcheck:
-      test: curl -s -o /dev/null -w "%{http_code}" "http://localhost/nginx-health" | grep -q "^200$" || false
+      test: |-
+        curl -s -o /dev/null -w "%{http_code}" --header 'Host: davidrunger.com' http://localhost/nginx-health | grep -q "^200$" || false
```
davidrunger, line 130: Sort of a side note: manually providing a `Host` header is a way to circumvent our efforts to require requests to go through Cloudflare (via DNS resolution), rather than permitting requests made directly to the server's IP address. However, we can circumvent that protection, as seen here, and hackers can do this, too (and, from some log lines that I have seen, I believe that they actually have done this and been noticed by me, which somewhat surprises me).

davidrunger:
Deploys are currently failing ([latest](https://github.com/davidrunger/david_runger/actions/runs/12858155508/job/35847011341)) due to the `nginx` healthcheck failing. This change should fix the healthcheck and, thus, deployments.

PR #5890, opened 2025-01-21
Don't require certbot, nginx healthchecks for deploy; tweak web check

davidrunger:
Our CPU is nowadays has a higher baseline load than I would like. I think that, partially as a result of this, the median response time in the past day has been ~180 ms (which is much higher than normal), although memory pressure and swap usage is also probably a factor. Therefore, I want to reduce the CPU load created by healthchecks.

To this end, I want to increase all healthcheck intervals to at least 5m.

However, this risks breaking deploys, if we require a healthcheck for deployment, but it has an infrequent interval (and so cannot recover quickly). To avoid this issue, I am removing certbot and nginx from services that are required to have a healthy status during deployment.

Regarding certbot, it's not super vital that certbot be running at all times. If there are issues, then I can address them when I get an email from LetsEncrypt about the SSL certificates needing renewal.

Regarding nginx, we don't even restart the `nginx` service during deploys, and if there is an nginx issue then I expect to receive notice about it from UptimeRobot.

This change increases the `web` healthcheck `interval` from 12s to 5m. We recently decreased it from 1m to 12s in #5865 (and increased the timeout from 1s to 3s), but I don't know that we can afford that amount of CPU usage. This change goes back past 1m all the way to 5m. To compensate (i.e. to hopefully make the healthcheck really unlikely to fail when it does run at these relatively infrequent intervals), we'll further increase the timeout to 5s.

Also, for the web rollout, we'll now curl the `/up` endpoint instead of the root. I think that the root endpoint might be slower than `/up` during a web rollout, because there won't be any prerenders yet, and, so, when making requests to `/`, we will be making requests to S3 looking for prerenders, which will never be there. `/up` won't have this issue.

Now, all `interval`s are at least 5m, which should not create too much CPU load.

### DIFF
--- a/bin/server/roll-out-web.sh
+++ b/bin/server/roll-out-web.sh
@@ -35,7 +35,7 @@ echo_iso8601 "Curling new container IP until success."
 attempt=1
 max_attempts=30
 delay=2
-url="http://$new_container_ip:3000/"
+url="http://$new_container_ip:3000/up"
 
 while [ $attempt -le $max_attempts ] ; do
   echo_iso8601 "attempt=$attempt max_attempts=$max_attempts url=$url"
--- a/bin/server/verify-expected-services.sh
+++ b/bin/server/verify-expected-services.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail # exit on any error, don't allow undefined variables, pipes don't swallow errors
 
-expected_running_services=(certbot clock nginx postgres redis-app redis-cache web worker)
+expected_running_services=(clock postgres redis-app redis-cache web worker)
 expected_services_not_running=()
 max_retries=30
 retry_delay=2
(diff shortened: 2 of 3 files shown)


davidrunger:
The `htop` graph I just watched is looking pretty good:

image

The Grafana graph is not looking much better, though:

image

Part of the time post-deploy I was looking at Grafana (which might have caused some non-trivial load), but part of the time I was not, and things don't look much better even when I wasn't looking at it.

Looking at `docker stats`, things seem reasonable-ish, given our server resource constraints and the number of services that I am trying to run. There are spikes in CPU usage every once in a while for various services, but nothing is steady at more than 0.x%. The "MEM USAGE" usage of e.g. `web` is very low, which I think reflects that most of the memory has probably been put into swap, which can't be good for performance. At the end of the day, the main thing is that I need more server memory, if I'm going to be running all of these services.

PR #5891, opened 2025-01-22
[docker-compose] Specify memory reservations for key services

davidrunger:
In this change, I am specifying memory reservations for all services that are actually potentially end-user-impacting: clock, nginx, postgres, redis-app, redis-cache, web, and worker.

To determine the memory reservation for each service, I started by looking at the mean total memory consumption (RAM + swap) and the RSS memory (i.e. RAM utilization) of that service over the past 24 hours (from the "Memory Usage Details" panel of the "Docker monitoring" dashboard). NOTE: That graph shows values in MiB, but in the table below I have converted them to MB, which is also the unit for the Docker Compose memory reservation.

```
             Total   RSS     Reservation (added in this change)
clock        1.17    0.35    6
nginx        4.35    1.18    10
postgres     39.43   2.46    50
redis-app    8.05    1.43    12
redis-cache  2.71    0.56    6
web          226.5   63.44   230
worker       187.7   52.74   36
```

In almost all cases, we are adding a little padding above the total mean memory usage (including swap). This will hopefully ensure that, for these services, usage of swap will be fairly minimal. In the case of clock and redis-cache, we are forced to add extra padding, because the minimum allowed reservation size is 6 MB. In some cases, like nginx, we are also adding a little extra padding, because the bang for buck (just a few extra MB of reserved memory) seems very good, in terms of making sure that NGINX can run fully in RAM.

The notable exception is worker, which wants to use ~188 MB of memory, but which we are only reserving 36 MB for. This is because worker's performance is not very critical. It tries to use significantly more memory than it is worth paying for. So, we are only going to reserve about 19% of the memory that it would like to use.

The total of the reservations is 350 MB. I'm not sure that we have a whole lot more than that available for use by our Docker services, after Debian's memory usage is accounted for.

Since we aren't specifying any memory limits or reservations for other services, I guess that this change will leave them/Docker to sort out how to allocate memory among them? I guess that I am going to deploy this, and look at the aforementioned graph to see how the memory usage of all of the Docker Compose services changes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -78,6 +78,10 @@ services:
   clock:
     <<: *default-rails-config
     command: ['bin/skedjewel']
+    deploy:
+      resources:
+        reservations:
+          memory: 6M
     healthcheck:
       test: ps -eo cmd | grep -P '^[b]in/skedjewel$'
       start_period: 10s
@@ -122,6 +126,10 @@ services:
     depends_on:
       vector:
         condition: service_started
+    deploy:
+      resources:
+        reservations:
+          memory: 10M
     entrypoint: '/bin/sh -c ''while :; do sleep 6h & wait $${!}; nginx -s reload; done & /docker-entrypoint.sh nginx -g "daemon off;"'''
     env_file:
       - .env.nginx.local
@@ -158,6 +166,10 @@ services:
     depends_on:
       vector:
         condition: service_started
+    deploy:
+      resources:
+        reservations:
+          memory: 50M
     env_file:
       - .env.postgres.local
     healthcheck:
@@ -202,6 +214,10 @@ services:
       # - The ${variable:?message} syntax causes shell to exit with a non-zero
       #   code and print a message, when the variable is not set or empty
       - redis-server --requirepass "$${REDIS_PASSWORD:?REDIS_PASSWORD is required}"
+    deploy:
+      resources:
+        reservations:
+          memory: 12M
     env_file:
       - .env.redis-app.local
     volumes:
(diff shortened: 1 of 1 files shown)


davidrunger:
#5891 (this PR) and #5892 have decreased memory usage on the server (deploys around 10pm and 12am):

[Graph 1](https://grafana.davidrunger.com/d/rYdddlPWk/node-exporter-full?orgId=1&from=2025-01-22T00:41:48.426Z&to=2025-01-22T06:41:48.426Z&timezone=browser&var-datasource=beafcdfad4d1cc&var-job=node&var-node=node_exporter:9100&var-diskdevices=%5Ba-z%5D%2B%7Cnvme%5B0-9%5D%2Bn%5B0-9%5D%2B%7Cmmcblk%5B0-9%5D%2B&refresh=10s&viewPanel=panel-78):

image

[Graph 2](https://grafana.davidrunger.com/d/m0arCBf7k/docker-monitoring?orgId=1&from=2025-01-22T00:40:51.524Z&to=2025-01-22T06:40:51.524Z&timezone=browser&var-job=$__all&var-service=$__all&var-node=$__all&var-container=$__all&refresh=10s&viewPanel=panel-12):

(Note that an increased fraction of Docker services memory is now in RSS rather than swap.)

image

[Graph 3](https://grafana.davidrunger.com/d/m0arCBf7k/docker-monitoring?orgId=1&from=2025-01-22T00:44:20.959Z&to=2025-01-22T06:44:20.959Z&timezone=browser&var-job=$__all&var-service=$__all&var-node=$__all&var-container=$__all&refresh=10s&viewPanel=panel-11):

(I tried to select a few key services to show.)

image

davidrunger:
Memory usage is _way_ down after a reboot ([link](https://grafana.davidrunger.com/d/rYdddlPWk/node-exporter-full?orgId=1&from=2025-01-21T15:28:03.503Z&to=2025-01-22T15:28:03.504Z&timezone=browser&var-datasource=beafcdfad4d1cc&var-job=node&var-node=node_exporter:9100&var-diskdevices=%5Ba-z%5D%2B%7Cnvme%5B0-9%5D%2Bn%5B0-9%5D%2B%7Cmmcblk%5B0-9%5D%2B&refresh=10s&viewPanel=panel-78)):

image

I don't know how much of that memory use drop would have occurred without these changes vs how much is a result of the reboot with these changes in place.

PR #5898, opened 2025-01-22
Stop sending logs to Papertrail [DEV-139]

davidrunger:
I don't think I really am going to get much value out of Papertrail, now that I have my own logging set up via Vector and Loki, viewable via Grafana.

This might save some CPU and/or memory in Vector, and it is probably better not to be sending data out to an external service, all other things being equal.

A Papertrail-like search experience can be had by going to "Explore" in Grafana, selecting Loki as the data source, using a query like `{service_name=~".+"} |~ "your_query_here"`, and clicking the "Live" button.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -280,11 +280,8 @@ services:
       - '--require-healthy'
       - 'true'
       - '--watch-config'
-    env_file:
-      - .env.vector.local
     image: timberio/vector:latest-alpine
     networks:
-      - external
       - internal
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock:ro
--- a/vector/vector.yaml
+++ b/vector/vector.yaml
@@ -1,5 +1,4 @@
 sources:
-  # Read Docker container logs
   docker:
     type: docker_logs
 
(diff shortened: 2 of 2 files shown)


davidrunger:
Vector memory use is down a bit: [link](https://grafana.davidrunger.com/d/m0arCBf7k/docker-monitoring?orgId=1&from=2025-01-22T18:49:58.829Z&to=2025-01-22T19:49:58.830Z&timezone=browser&var-job=$__all&var-service=$__all&var-node=$__all&var-container=$__all&refresh=20s&viewPanel=panel-11)

(However, this might possibly be due more to the fact that I restarted Vector, rather than being due to the removal of Papertrail.)

image

CPU looks like it might be down a hair, but harder to tell: [link](https://grafana.davidrunger.com/d/m0arCBf7k/docker-monitoring?orgId=1&from=2025-01-22T18:50:58.837Z&to=2025-01-22T19:50:58.837Z&timezone=browser&var-job=$__all&var-service=$__all&var-node=$__all&var-container=$__all&refresh=20s&viewPanel=panel-2)

image

PR #5904, opened 2025-01-23
[CheckLinks] Space LaunchLinkChecksForPage jobs 5 minutes apart

davidrunger:
Motivation: avoid a burst of many `CheckLinks::Checker` jobs running all around the same time right after the `CheckLinks::LaunchLinkChecksForPage` jobs are run. Previously, the `CheckLinks::LaunchLinkChecksForPage` jobs were run without any spacing between them. Each `CheckLinks::LaunchLinkChecksForPage` job, in turn, would launch many `CheckLinks::Checker` jobs, which were spaced with respect to any other `CheckLinks::Checker` jobs launched by each individual `CheckLinks::LaunchLinkChecksForPage` job, but these jobs were being launched by `CheckLinks::LaunchLinkChecksForPage` jobs that were themselves _not_ spaced. So, right after the `CheckLinks::LaunchLinkChecksForPage` jobs ran more or less immediately, then there would be a burst of many `CheckLinks::Checker` jobs running all around the same time. By spacing the `CheckLinks::LaunchLinkChecksForPage` jobs, as well, we should avoid this burst.

[Graph](https://grafana.davidrunger.com/goto/cXEzDKONg?orgId=1):

image

^ There is a large number of `CheckLinks::Checker` jobs running right away, because approximately all of the webpages for which we are checking the links have at least one link. As time goes by, there is less and less bunching / fewer jobs being run ~simultaneously, because not so many websites have many links. The long tail to the right comes from the davidrunger.com homepage, which has very many links.

### DIFF
--- a/app/workers/check_links/launch_page_fetches.rb
+++ b/app/workers/check_links/launch_page_fetches.rb
@@ -4,9 +4,11 @@ class CheckLinks::LaunchPageFetches
   SITEMAP_URL = 'https://davidrunger.com/sitemap.xml'
 
   def perform
-    sitemap_urls.each do |url|
-      CheckLinks::LaunchLinkChecksForPage.perform_async(url)
-    end
+    launch_with_spacing(
+      worker: CheckLinks::LaunchLinkChecksForPage,
+      arguments_list: sitemap_urls.map { [it] },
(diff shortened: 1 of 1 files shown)


davidrunger:
The jobs are much more spread out, now:

https://grafana.davidrunger.com/goto/DRYnXtONg?orgId=1

image

PR #5915, opened 2025-01-23
Bump eslint from 8.57.1 to 9.18.0 (& @vue/eslint-config-typescript)

davidrunger:
Done via `pnpm update eslint @vue/eslint-config-typescript` and other commands, plus manual changes.

I didn't really read through the migration documentation, and in general I'm not super sure what all is going on here. There definitely might be some things that could be improved here. But it seems to pass when things are good, and it seems to catch at least some violations.


.eslintignore
davidrunger, line 1: (This is now a section in the new `eslint.config.mjs` config file.)

.eslintrc.js
davidrunger, line 1: (Moved/translated to `eslint.config.mjs`.)

lib/test/tasks/run_eslint.rb
```diff
@@ -3,7 +3,7 @@ class Test::Tasks::RunEslint < Pallets::Task
 
   def run
     execute_system_command(<<~COMMAND)
-      ./node_modules/.bin/eslint --max-warnings 0 --ext .js,.ts,.vue app/javascript/
```
davidrunger, line 6: (These extension restrictions have been moved (in a different format) to `eslint.config.mjs`.)

PR #5923, opened 2025-01-25
Don't enqueue SaveRequest workers that won't save anyway [DEV-140]

davidrunger:
Background: We have been storing request data in Redis and running SaveRequest workers for requests with an uptime_robot query param and for blog#assets requests. However, this is wasteful, since we ultimately won't create Request records for these requests.

This change stops storing data for such requests in Redis and stops running SaveRequest workers for these requests. The motivation is essentially to avoid wasted Redis requests, logs, Sidekiq queue consumption, etc.

As part of this change, I have refactored some existing checks for certain other requests that we already have been proactively skipping (specifically, requests to HealthChecksController and AnonymousController). I have now rolled up all such checks into a unified PORO, which leverages pattern matching for what I think is a pretty clean way to describe the conditions in which we want to skip saving a Request record.

### DIFF
--- a/app/poros/save_request/skip_checker.rb
+++ b/app/poros/save_request/skip_checker.rb
@@ -1,0 +1,21 @@
+class SaveRequest::SkipChecker
+  def initialize(params:)
+    @controller = params['controller'].presence!
+    @action = params['action'].presence!
+    @params = params.presence!
+  end
+
+  def skip?
+    case [@controller, @action, @params.symbolize_keys]
+    in (
+      ['health_checks', _, _] |
+      ['anonymous', _, _] |
+      ['blog', 'assets', _] |
+      [_, _, { uptime_robot: _ }]
+    )
+      true
+    else
+      false
+    end
+  end
+end
--- a/config/initializers/monkeypatches/object.rb
+++ b/config/initializers/monkeypatches/object.rb
@@ -1,10 +1,12 @@
 module PresenceBangMonkeypatch
+  class ObjectNotPresent < StandardError ; end
+
   def presence!(error_message = nil)
     if present?
       self
     else
       error_message ||= "Expected object to be `present?` but was #{inspect}"
-      raise(error_message)
+      raise(ObjectNotPresent, error_message)
     end
   end
 end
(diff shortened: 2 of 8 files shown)


app/poros/save_request/skip_checker.rb
```diff
@@ -0,0 +1,21 @@
...
+    @params = params.presence!
+  end
+
+  def skip?
+    case [@controller, @action, @params.symbolize_keys]
+    in (
+      ['health_checks', _, _] |
+      ['anonymous', _, _] |
+      ['blog', 'assets', _] |
+      [_, _, { uptime_robot: _ }]
```
davidrunger, line 14: This might be one of the first times that I have leveraged Ruby pattern matching? I think it works pretty well here.

config/initializers/monkeypatches/object.rb
```diff
@@ -1,10 +1,12 @@
 module PresenceBangMonkeypatch
+  class ObjectNotPresent < StandardError ; end
```
davidrunger, line 2: Adding this error class was partially motivated by a desire to make it easier to test for this error to be raised, but I think it's probably also a good idea, in general, to have this be a dedicated error class (for better Rollbar tracking and other observability reasons).

app/poros/save_request/skip_checker.rb
```diff
@@ -0,0 +1,21 @@
+class SaveRequest::SkipChecker
```
davidrunger, line 1: I am feeling super happy with my decision to create `app/poros/`, where I can just throw things like this. In that sense, it's sort of a junk drawer, but I don't mind that. Many people might just call it `app/services/`, but I sort of like the low pressure and informality of "POROs". It makes me feel free to put in here whatever I think will be most useful.

spec/poros/save_request/skip_checker_spec.rb
```diff
@@ -0,0 +1,75 @@
+RSpec.describe SaveRequest::SkipChecker do
```
davidrunger, line 1: One of the other nice things about `app/poros/` is that they tend to be relatively easy to test, which I think is reflected in the pretty straightforward tests below.

PR #5931, opened 2025-01-27
[logs/text] Center log entries horizontally [LOG-29]

davidrunger:
Before / After :

image

### DIFF
--- a/app/javascript/logs/components/data_renderers/TextLog.vue
+++ b/app/javascript/logs/components/data_renderers/TextLog.vue
@@ -1,7 +1,7 @@
 <template lang="pug">
-.flex.justify-center
-  .container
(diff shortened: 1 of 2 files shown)


app/javascript/logs/components/data_renderers/TextLog.vue
```diff
@@ -1,7 +1,7 @@
 <template lang="pug">
-.flex.justify-center
-  .container
-    table.text-log-table
+div
+  .flex.justify-center
+    table.text-log-table.max-w-4xl
```
davidrunger, line 4: This is a bit narrower than what we had before. I want to try it out. Narrow can be nice, I think, because it means less horizontal eye movement.

PR #5946, opened 2025-01-28
[home] Use SVGs (not devicons font) for skill icons [HOME-10]

davidrunger:
The devicons font is generating a warning in the Firefox console about "Glyph bbox was incorrect". This annoys me. I want the console to be clean, especially on the homepage.

To fix this, I am migrating away from using the devicons font. Instead, we will now render the icons for my listed skills using SVGs (mostly obtained from svglogos.dev and the RSpec logo from svgrepo.com).

I did need to slightly modify the logos downloaded from svglogos.dev, because some of them had "linear gradients" where the `id` was like `linearGradient-1`, but other SVGs also had that some gradient, and so one would be selected at random, making the colors wrong for some of the logos. I fixed this by adding the file name into the `id`s (e.g. `linearGradient-jira-1`, thereby making them unique), using this command:

```bash
for file in app/assets/images/dev-icons/*.svg; do
  filename=$(basename "$file" .svg)
  sed -i -E "s/(\blinearGradient-)([0-9]+)/\1${filename}-\2/g" "$file"
done
```

Before (from Percy):

029675f05f476fd7a41f5dcdcc4e6b44c89ed1b99ead4bb7a98ceddde8013c4b

After (from Percy):

6761efcdb811e798019171a91a8e9647d1e3d071e3d1a89ff11d4b2553516c81


lib/test/tasks/run_file_size_checks.rb
```diff
@@ -17,7 +17,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'groceries*.css' => (108..118),
     'groceries*.js' => (441..451),
     'home*.css' => (0..10),
-    'home*.js' => (146..156),
+    'home*.js' => (239..249),
```
davidrunger, line 20: This is a significant increase, but we will save more bytes than this by virtue of no longer loading the devicons font. However, loading JavaScript is probably at least somewhat more costly, byte for byte, than loading a font. Overall, though, I think that the user-experienced rendering performance will actually probably be improved?, because I think that the SVGs will now be stored directly in the pre-rendered HTML pages, which I think will allow a user to see the skill icons more quickly than if their browser also had to download a font from a separate server, first.

spec/features/home_spec.rb
```diff
@@ -7,15 +7,6 @@
       Full stack web developer
     HEADLINE
 
-    # iff using Percy, make sure that devicon font has loaded, to keep screenshots consistent
-    if ENV.fetch('PERCY_BUILD_ID', nil).present?
-      wait_for do
-        page.evaluate_script('document.fonts.check("65px devicon")')
-      end.to eq(true)
```
davidrunger: Hopefully Percy will be reliable with these SVGs?, but I'm not confident that it will be.

app/views/home/index.html.haml
```diff
@@ -1,20 +1,2 @@
-- content_for(:extra_css) do
-  -# Always load `devicons` (in a nonblocking way (https://web.dev/defer-non-critical-css/)).
-  -# This modification complies with our CSP: https://stackoverflow.com/a/56400044/4009384
-  - devicon_href = 'https://cdn.jsdelivr.net/gh/devicons/devicon@v2.16.0/devicon.min.css'
-  = javascript_tag(nonce: true) do
```
davidrunger, line 5: It's nice to be able to remove this complex/ugly code.

bin/prerender
```diff
@@ -63,17 +63,8 @@ class Prerenderer
...
     @html_doc.to_s
   end
-
-  def remove_devicon_stylesheet!
-    @html_doc.
-      css(
-        'link[rel="stylesheet"][href*="devicon"][data-lazyloader-state="spent"]',
-      ).
-      remove
-  end
```
davidrunger, line 76: It's nice to be able to remove this, as well.

app/javascript/home/components/SkillRow.vue
```diff
@@ -21,42 +18,54 @@ const props = defineProps({
...
+    const fileName = props.name.toLowerCase().replace(/[^a-z0-9]+/g, '-');
+    const filePath = `../assets/images/dev-icons/${fileName}.svg`;
+
+    const svgFileForPath = svgFiles[filePath];
+
+    if (svgFileForPath) {
+      const fileSvgContent = await svgFileForPath();
+
+      if (typeof fileSvgContent === 'string') {
+        svgContent.value = fileSvgContent;
```
davidrunger, line 41: I feel like I am trying to do something pretty simple -- inline some SVGs into the HTML -- but I guess it is not straightforward to do that with Vite.

PR #5973, opened 2025-01-31
Use propshaft for asset pipeline (and remove sprockets) [DEV-147]

davidrunger:
Sprockets is deprecated as of Rails 8. The new asset pipeline gem is propshaft. To stay current, this change remove sprockets (and sprockets-rails and sassc) and replaces it with propshaft.

Propshaft, however, does not perform any transformations (e.g. Sass to CSS) -- and ActiveAdmin 3's stylesheets are written in CSS. Therefore, in order to continue to use the ActiveAdmin stylesheets, we will now compile them via Vite rather than via Sprockets. This excellent post explained how to do that: https://github.com/activeadmin/activeadmin/discussions/7947#discussioncomment-5867902 .

Unfortunately, there are lots of Sass deprecation warnings about the ActiveAdmin Sass stylesheets. I guess that I will just ignore them, for now. To that end, in this change, I am suppressing the output of the admin assets compilation step in the `CompileJavaScript` test step.

(Side note: I think that ActiveAdmin 4 might be moving from Sass to pure CSS?, in which case we might be able to remove the Vite compilation of ActiveAdmin stylesheets.)

It was important to specify type="module" for the admin scripts so that there aren't variable name conflicts between modules when multiple are rendered on a page (e.g. how on the asset sizes admin page we include both the active_admin script and the charts script).


app/javascript/admin_entrypoints/charts.ts
```diff
@@ -3,8 +3,4 @@ import Chartkick from 'chartkick';
 
 import 'chartjs-adapter-luxon';
 
-// @ts-expect-error NOTE: Importing Chartkick seems to overwrite `window.$` with
-// Chartkick's `Chart` object, which is minified as `$`? Undo this.
-window.$ = window.jQuery;
```
davidrunger, line 8: I think we don't need this anymore because we are doing `type="module"` for the scripts now (which I think keeps variables separate from each other within the scripts, even if the variables have the same names).

spec/spec_helper.rb
```diff
@@ -61,11 +61,14 @@
 OmniAuth.config.test_mode = true
 
 Capybara.register_driver(:cuprite) do |app|
+  timeout = is_ci ? 10 : 999_999
```
davidrunger, line 64: This allows for a debugging session with headful Chrome that won't be interrupted very quickly by Chrome being closed.

spec/support/cuprite_logger.rb
```diff
@@ -42,12 +42,12 @@ def puts(message)
...
           join("\n")
 
-      full_stack_trace = parsed_json.dig(
+      full_stack_trace = (parsed_json.dig(
         'params',
         'exceptionDetails',
         'stackTrace',
         'callFrames',
-      ).map { _1.values_at('functionName', 'url') }
+      ) || []).map { _1.values_at('functionName', 'url') }
```
davidrunger, line 50: This fixes an issue where some JavaScript exceptions did not have stack traces.

PR #5982, opened 2025-02-01
Disable external requests from browser in feature specs [DEV-149]

davidrunger:
... by default, but allowlisted domains may be specified for specific spec(s) via RSpec metadata.

**Motivation:** I believe that generally our specs should only test our own system and probably shouldn't rely (unless, perhaps, explicitly specified) on anything else on the Internet being in any certain state / behaving in any certain way.

Additionally, loading resources from the external Internet in specs might carry a nontrivial performance cost, and this change will prevent such situations from creeping into our test suite without notice.

I'm adding a test to check that external domains are blocked, but I am not adding a test to verify that they can be allowlisted (since such a test would require accessing an external domain, which would be ironically somewhat contrary to the purpose of this change, in the first place).

I moved some local variables in `spec_helper.rb` as methods in a `SpecHelper` module, so that they can be accessed from other files (e.g. `spec/support/cuprite/domain_restricted_driver.rb`), allowing for some modularization of the `spec_helper.rb` file, which otherwise will probably tend to sprawl somewhat.


config/initializers/content_security_policy.rb
```diff
@@ -6,7 +6,7 @@
 
 Rails.application.configure do
   # Allow disabling CSP in development because it breaks Vue devtools in Firefox.
-  if Rails.env.development? && ENV.key?('DISABLE_CSP')
+  if Rails.env.local? && ENV.key?('DISABLE_CSP')
```
davidrunger, line 9: I wanted to test that AJAX requests were also subjected to the same domain restrictions as the top level page navigations. (They are.) This required disabling the CSP. This change makes that easier.

spec/spec_helper.rb
```diff
@@ -1,6 +1,14 @@
...
 ENV['RAILS_ENV'] ||= 'test'
-is_ci = (ENV.fetch('CI', nil) == 'true')
+
+module SpecHelper
+  class << self
+    def is_ci? = ENV.fetch('CI', nil) == 'true'
+
+    def use_headful_chrome? = ENV.fetch('HEADFUL_CHROME', nil).present?
+  end
+end
```
davidrunger, line 10: Putting these as methods in a module (rather than local variables) makes it possible to access them from other files.

davidrunger: I think that this is the first time that I have used Ruby's relatively newly added one-line endless method definition style.

spec/spec_helper.rb
```diff
@@ -27,24 +37,24 @@
...
 require 'capybara/rails'
 require 'capybara/rspec'
 require 'capybara/email/rspec'
-require 'capybara-screenshot/rspec' unless use_headful_chrome
+require 'capybara-screenshot/rspec' unless SpecHelper.use_headful_chrome?
 require 'mail'
 require 'percy/capybara'
 require 'super_diff/rspec-rails'
+require Rails.root.join('spec/support/fixture_builder.rb').to_s
+Dir['spec/support/**/*.rb'].each { |file| require Rails.root.join(file) }
```
davidrunger, line 57: I wanted to move these requires down just because it seems better to have all of the gems loaded before we load our support files, in case our support files want to reference something from one of the gems.

spec/spec_helper.rb
```diff
@@ -223,6 +223,14 @@
     end
   end
 
+  # Permit allowing requests to specified external domains (which we block by
+  # default; see url_whitelist in Cuprite::DomainRestrictedDriver).
+  config.before do |example|
+    (example.metadata[:allowed_domains] || []).each do |allowed_domain|
+      page.driver.browser.url_allowlist << %r{\Ahttps://#{allowed_domain}/}
```
davidrunger: I tested and found that the `page.driver.browser.url_allowlist` gets reset with each spec. In other words, an allowed domain for one example will not continue to be allowed when a subsequent example is executed. (This is good.)

spec/support/cuprite/domain_restricted_driver.rb
```diff
@@ -0,0 +1,19 @@
+module Cuprite::DomainRestrictedDriver
```
davidrunger: It would maybe be somewhat better to have just had the `Capybara.register_driver(:domain_restricted_cuprite)` call at the top level of this file, but Zeitwerk complains about this. Putting it as a static module method (where the module name matches the file path) satisfies Zeitwerk.

spec/support/cuprite/browser_logger.rb
```diff
@@ -2,7 +2,7 @@
 
 using Rainbow
 
-class CupriteLogger
+class Cuprite::BrowserLogger
```
davidrunger, line 5: Since I already had this Cuprite-related class, and since I was going to create another Cuprite-related file (to register the domain-restricted driver), I wanted to create a `spec/support/cuprite/` directory, within which both could be organized. So that's why this constant name has changed. I also felt that `BrowserLogger` is likely than `Logger` to possibly conflict with a constant that might already be within the gem's `Cuprite` namespace.

spec/features/user_google_login_spec.rb
```diff
@@ -54,7 +54,7 @@
     context 'when Google responds with "This is Google OAuth."' do
       let(:google_response_content) { 'This is Google OAuth.' }
 
-      around do |spec|
+      before do
```
davidrunger, line 57: Changing from `around` to `before` is actually significant. Without this change, the test fails. The reason is because we have a global RSpec `before` block that checks which driver to use (depending on metadata), and we need that block to run before this block. When both are `before`s, that happens. When this one was an `around`, I think it was running first, which was breakingly problematic.

spec/support/cuprite/custom_drivers.rb
```diff
@@ -0,0 +1,38 @@
+module Cuprite::CustomDrivers
+  DOMAIN_RESTRICTED_CUPRITE = :domain_restricted_cuprite
+  UNRESTRICTED_CUPRITE = :unrestricted_cuprite
+
+  class << self
+    def register_with_capybara
+      register_driver(
+        DOMAIN_RESTRICTED_CUPRITE,
+        -> { { url_whitelist: %r{\Ahttp://localhost:#{Capybara.server_port}/} } },
```
davidrunger, line 9: This needs to be wrapped in a proc to delay its execution (I guess until we are within the `Capybara.register_driver` block), because otherwise `Capybara.server_port` is not yet defined, and so basically nothing gets allowlisted and pretty much every spec fails.

PR #6003, opened 2025-02-04
[specs/feature] Write server log to file on spec failure [DEV-156]

davidrunger:
This should be helpful to debug failing or (especially) flaking specs.

I ran the test suite 3 times on `main` and then 3 times on this branch. Results:

== main
1:29.49 total
1:23.02 total
1:20.45 total
Average: 1:24.3

== logs on failure
1:21.00 total
1:24.01 total
1:14.76 total
Average: 1:19.9

So, in those numbers, there's a 4.4 second speedup on this branch. I can't think of how/why this branch would really provide a speedup vs main, but the takeaway is that there does not appear to be a significant slowdown, at least.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -44,29 +44,37 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 10 # this might cause issues if there are more than 10 commits in a PR (?)
+
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
           bundler-cache: true
+
       - uses: actions/setup-node@v4
         with:
           node-version-file: .node-version
+
       - name: Install pnpm
         run: npm install -g pnpm@9
+
       - name: Set pnpm store directory
         run: echo "PNPM_STORE_PATH=$(pnpm store path)" >> $GITHUB_ENV
+
       - uses: actions/cache@v4
         name: Set up pnpm cache
         with:
           path: ${{ env.PNPM_STORE_PATH }}
           key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
           restore-keys: ${{ runner.os }}-pnpm-store-
+
       - name: Run ShellCheck
         uses: ludeeus/action-shellcheck@master
         with:
           scandir: './bin'
+
       - name: Set GIT_REV environment variable
         uses: ./.github/actions/git_rev
+
       - name: Run tests & linters
         run: bin/run-tests
         env:
@@ -88,6 +96,15 @@ jobs:
           PERCY_TOKEN: '${{secrets.PERCY_TOKEN}}'
           CODECOV_TOKEN: '${{secrets.CODECOV_TOKEN}}'
           RAILS_MASTER_KEY: '${{secrets.RAILS_MASTER_KEY}}'
+
+      - name: Save failed feature spec logs
+        uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: failed_feature_spec_logs
+          path: log/failed_feature_specs/
+          retention-days: 5
+
       - name: Save fixtures artifact
         uses: actions/upload-artifact@v4
         if: failure()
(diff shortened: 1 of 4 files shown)


config/environments/test.rb
```diff
@@ -54,15 +54,16 @@
   # Raise exceptions on deprecation notices.
   config.active_support.deprecation = :raise
 
-  if ENV.fetch('TEST_LOGGING', nil) == '1'
-    # :nocov:
-    config.log_level = :debug
-    config.active_record.verbose_query_logs = true
+  config.active_record.verbose_query_logs = true
```
davidrunger, line 57: I'm setting `verbose_query_logs = true` always, because we don't write the logs at all in the default condition (when `TEST_LOGGING` is not `1`), so presumably it doesn't negatively impact performance then? And when I do have logging on, then I think I will always want the verbose query logs.

.github/workflows/ruby.yml
```diff
@@ -44,29 +44,37 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 10 # this might cause issues if there are more than 10 commits in a PR (?)
+
```
davidrunger, line 47: (Unrelated change; I think adding newlines between the steps makes it easier to parse this list.)

spec/spec_helper.rb
```diff
@@ -194,6 +194,40 @@ def use_headful_chrome? = ENV.fetch('HEADFUL_CHROME', nil).present?
     RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length = 2_000
   end
 
+  config.around(:each, type: :feature) do |example|
+    $test_log_string_io ||= StringIO.new
```
davidrunger, line 198: Unfortunately, it seems that this needs to be a global variable. (Or, at least, I didn't want to invest in finding a way to make this work without using a global variable.) Otherwise, what seems to happen is that only the first spec that is executed is able to write ActionController logs to the StringIO. If we try to do this as a local variable, then the later examples create a new StringIO, but I think that the Puma test server remains booted with a reference to the StringIO of the first example, and so, Puma doesn't write anything to the local StringIO that we might try to use for all examples other than the first one executed.

spec/spec_helper.rb
```diff
@@ -194,6 +194,40 @@ def use_headful_chrome? = ENV.fetch('HEADFUL_CHROME', nil).present?
...
+    $test_log_string_io.reopen
+
+    string_io_logger = ActiveSupport::Logger.new($test_log_string_io)
+    string_io_logger.formatter = Rails.logger.formatter.dup
+    string_io_logger.level = :debug
+
+    Rails.with(logger: string_io_logger) do
+      Sidekiq.default_configuration.with(logger: string_io_logger) do
+        # We can't just do ActiveRecord::Base.with(...) because `with` is an ActiveRecord method.
+        Object.instance_method(:with).bind_call(ActiveRecord::Base, logger: string_io_logger) do
```
davidrunger: For some Rails stuff (like ActionCable and ActionController), they seem to use the `Rails.logger`, and so changing that causes them to log to our StringIO. However, Sidekiq and ActiveRecord seem to create their own loggers (maybe initially based partially on `Rails.logger`, I'm not sure), and so changing `Rails.logger` doesn't affect their behavior. That's why here we are also adjusting the Sidekiq and ActiveRecord loggers, specifically.

spec/spec_helper.rb
```diff
@@ -194,6 +194,40 @@ def use_headful_chrome? = ENV.fetch('HEADFUL_CHROME', nil).present?
...
+      end
+    end
+
+    if example.exception
+      # Prepare file name and directory
+      timestamp = example.execution_result.started_at.iso8601
+      description_as_brief_file_name =
+        example.full_description.
+          parameterize.
+          last(50).
```
davidrunger, line 220: I'm limiting the number of characters because I think that some systems will complain about file names being too long (though I forget which and don't know what the specific limit might be). Also, a too-long filename begins to be unwieldy, anyhow.

PR #6018, opened 2025-02-05
[admin/deploys] Add GitHub compare view link since previous [ADM-2]

davidrunger:
Shoutout to ChatGPT for giving me this `LAG` query! (Private link: https://chatgpt.com/c/67a36fb6-74c0-8007-99c8-67f662464c79)

### DIFF
--- a/spec/support/fixture_builder.rb
+++ b/spec/support/fixture_builder.rb
@@ -121,6 +121,7 @@ FixtureBuilder.configure do |fbuilder|
 
     # Deploys
     name(:deploy, create(:deploy))
+    create(:deploy)
 
     # Events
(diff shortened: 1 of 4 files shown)


spec/support/fixture_builder.rb
```diff
@@ -121,6 +121,7 @@
 
     # Deploys
     name(:deploy, create(:deploy))
+    create(:deploy)
```
davidrunger, line 124: We are adding a [spec](https://github.com/davidrunger/david_runger/pull/6018/files#diff-4399d8edfb259ed764255b47aa8ff01a2cba4810562b824124a69e57915ff451R5-R10) that makes sure we have at least two deploys. We might as well create them upfront here in the fixtures.

PR #6028, opened 2025-02-06
Bump postgres from 16.3 to 17.2 [DEV-113]

davidrunger:
Also, add script documenting upgrade plan.

Also, change the name of the volume that we are attaching to the postgres service. This will give us a "fresh start", which we sort of need, because otherwise Postgres will complain/error that the data was created with Postgres 16 but we are now running on Postgres 17. The data will be copied from the old volume to the new volume by dumping a backup from version 16 and then restoring from that backup after we are on version 17.

NOTE: The `docs/postgres-16-to-17-upgrade.sh` script being added here is not for execution as a script. Rather, this is a playbook of commands that I plan to manually copy-paste for execution on the server. I am adding the script for documentation (hence why it is in `docs/`) and so that hopefully I can reuse it (maybe with some tweaks) when doing future Postgres upgrades.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -199,13 +199,13 @@ services:
       interval: 5m
       timeout: 1s
       retries: 1
-    image: postgres:16.3-alpine
+    image: postgres:17.2-alpine
     logging: *default-logging
     memswap_limit: 99G
     networks:
       - internal
     volumes:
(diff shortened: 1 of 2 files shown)


davidrunger:
I am happy to report that everything went pretty smoothly! We had just a few minutes of downtime.

PR #6045, opened 2025-02-07
Track scrolling with http/ky/fetch (not sendBeacon) [HOME-15]

davidrunger:
At least one ad blocker (uBlock Origin) blocks `sendBeacon`, but I doubt that any block `fetch`. We use `sendBeacon` for external link clicks because those need to be sent asynchronously (because otherwise the request might be cancelled as the user navigates away), but we don't really need that for scrolls. So, this change uses `fetch` (via ky, via our http wrapper library) to track scrolls, which should be more reliable.

### DIFF
--- a/config/initializers/wrap_parameters.rb
+++ b/config/initializers/wrap_parameters.rb
@@ -5,7 +5,7 @@
 
 # Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
 ActiveSupport.on_load(:action_controller) do
-  wrap_parameters format: [:json]
+  wrap_parameters format: []
   ActionController::Parameters.action_on_unpermitted_parameters = false
 end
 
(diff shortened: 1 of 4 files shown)


config/initializers/wrap_parameters.rb
```diff
@@ -5,7 +5,7 @@
 
 # Enable parameter wrapping for JSON. You can disable this by setting :format to an empty array.
 ActiveSupport.on_load(:action_controller) do
-  wrap_parameters format: [:json]
+  wrap_parameters format: []
```
davidrunger, line 8: (Basically unrelated, but I noticed while looking at these requests in the Rails server logs that `wrap_parameters` was enabled, which I sort of hate, so I'm disabling it here.)

lib/test/tasks/run_file_size_checks.rb
```diff
@@ -17,7 +17,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'groceries*.css' => (108..118),
     'groceries*.js' => (441..451),
     'home*.css' => (0..10),
-    'home*.js' => (239..249),
+    'home*.js' => (260..270),
```
davidrunger, line 20: It's sort of sad to be increasing the page weight by ~21 KiB with this change, but I am pretty interested in tracking these scroll events, so I think it's worth it.

PR #6047, opened 2025-02-08
Move toasting of validation errors from http.ts to groceries/store.ts

davidrunger:
Motivation: lower page weight for apps (like home) that don't need this toasting functionality.

### DIFF
--- a/app/javascript/groceries/components/CheckInItem.vue
+++ b/app/javascript/groceries/components/CheckInItem.vue
@@ -15,10 +15,11 @@ li.flex.items-center.break-word.mb-2(
       span(v-if='item.needed > 1') {{' '}} ({{item.needed}})
     span {{ ' ' }}
     ElTooltip(
+      v-if="isSpouseItem(item)"
       content="Spouse item"
       placement="top"
     )
-      HeartFilledIcon.text-red-500(v-if="isSpouseItem(item)")
+      HeartFilledIcon.text-red-500
     span {{ ' ' }}
     el-button(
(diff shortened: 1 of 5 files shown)


app/javascript/groceries/components/CheckInItem.vue
```diff
@@ -15,10 +15,11 @@ li.flex.items-center.break-word.mb-2(
       span(v-if='item.needed > 1') {{' '}} ({{item.needed}})
     span {{ ' ' }}
     ElTooltip(
+      v-if="isSpouseItem(item)"
```
davidrunger, line 18: (Unrelated change. Noticed an issue when running the groceries feature specs. See commit: https://github.com/davidrunger/david_runger/pull/6047/commits/a0fa300045e9963d3ed8b4e5a9f4d426b2c77356 .)

lib/test/tasks/run_file_size_checks.rb
```diff
@@ -17,7 +17,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'groceries*.css' => (108..118),
     'groceries*.js' => (441..451),
     'home*.css' => (0..10),
-    'home*.js' => (260..270),
+    'home*.js' => (252..262),
```
davidrunger, line 20: Reducing the page weight of the home app was a primary motivation for this change, so it's nice to see it paying off here (even if only in the form of ~8 KiB saved -- every bit counts).

PR #6063, opened 2025-02-09
Split feature specs into three separate tasks [DEV-126]

davidrunger:
This is quite rough/ugly, but I'm interested to see if it can significantly improve test run time. I think that it might, in which case I think that I might go forward with this, despite the significant ugliness and to some extent risk of brittleness.

I guess that splitting into three feature spec groups makes sense because currently we are running specs with a pallets concurrency of 3, so no more than that many feature specs can run simultaneously, anyhow.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -49,7 +49,10 @@ Rails.application.configure do
 
   # Set host to be used by links generated in mailer templates.
   # https://github.com/DavyJonesLocker/capybara-email#setting-your-test-host
-  config.action_mailer.default_url_options = { host: 'localhost', port: 3001 }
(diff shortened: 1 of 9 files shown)


davidrunger:
I think this isn't worth it.

For one thing, it's not playing well with Percy.

For another thing, although it greatly increases parallelism, it also greatly increases CPU time, for some reason, to the point that it doesn't really save any time (i.e. the blue line below for WallClockTime stays flat). I wish I had a better sense of why that might be. Alas, though, I don't.

image

This sort of makes me think that I might not have much success increasing parallelism, even with a more sophisticated approach. It seems that the CI machines do not handle running multiple feature specs simultaneously very well, for some reason.

PR #6086, opened 2025-02-10
[github-actions] Print number of CPUs

davidrunger:
I think GitHub claims that there will be 4. I want to verify that this is the case.


davidrunger:
Supposedly there are indeed 4: https://github.com/davidrunger/david_runger/actions/runs/13250458497/job/36986948844?pr=6086

image

Closing.

PR #6089, opened 2025-02-11
Tweak CSS for text log entries, p tags, li tags

davidrunger:
It can be pretty hard to describe the motivation for CSS changes, so I'm not even really going to try. I will make at least one comment on the PR, though.

### DIFF
--- a/app/javascript/logs/components/data_renderers/TextLog.vue
+++ b/app/javascript/logs/components/data_renderers/TextLog.vue
@@ -58,7 +58,7 @@ ol {
 
   li {
     display: block;
-    margin-bottom: -13px;
+    margin-top: 0.5em;
     margin-left: $li-indent;
 
     &::before {
@@ -88,7 +88,7 @@ table.text-log-table {
   }
 
   td {
-    padding: 5px 10px;
+    padding: 18px 10px;
     word-break: initial;
     white-space: pre-wrap;
     line-height: 1.13rem;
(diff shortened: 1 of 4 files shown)


app/javascript/logs/components/data_renderers/TextLog.vue
```diff
@@ -104,9 +104,13 @@ table.text-log-table {
...
       &:not(:first-of-type) {
-        margin-top: 10px;
+        margin-top: 16px;
       }
     }
   }
+
+  td:has(ol, li) {
+    white-space: unset;
+  }
```
davidrunger, line 114: This is sort of a compromise solution.

Above, we have `white-space: pre-wrap;` for `td`s. This is good when a user just wants to enter some notes separated by a single line, whereas for Markdown to treat them as separate paragraphs (rather than just condensing the whitespace down to spaces on a single rendered line) the user would technically need to insert two newlines, rather than just the one.

```
Users like typing less.
`white-space: pre-wrap` let them do it.
Rendering as separate lines, even with just one newline between each line.
```

However, this looks terrible if/when there is a Markdown-parsed `ul` or `ol` (I guess because the combination of vertical spacing from the ul/ol/li elements plus the vertical spacing resulting from the user's entered newlines creates way too much vertical spacing). Therefore, here, we `unset` the `white-space` property at least for `td`s that have an `ol` and/or an `li`.

PR #6107, opened 2025-02-12
Split feature specs into two separate tasks [DEV-171]

davidrunger:
This is quite rough/ugly, but I'm interested to see if it can significantly improve test run time. I think that it might, in which case I think that I might go forward with this, despite the significant ugliness and to some extent risk of brittleness.

I previously tried splitting them into _three_ separate tasks ( #6063 ), but that didn't seem to help much.

However, I'm wondering if that's just because, between the server and the RSpec process and the Chrome process, we were just trying to use too much concurrency. But maybe two separate feature spec tasks running concurrently will be faster than either one or three? This change tries that.

Percy documentation about parallel builds related to some of these changes: https://www.browserstack.com/docs/percy/integrate/parallel-test-suites#parallel-tests-on-the-same-machine


lib/test/requirements_resolver.rb
```diff
@@ -33,22 +35,35 @@ def dependency_map
         Test::Tasks::RunVitest => Test::Tasks::PnpmInstall,
         Test::Tasks::RunAnnotate => Test::Tasks::SetupDb,
         Test::Tasks::RunTypelizer => Test::Tasks::BuildFixtures,
-        Test::Tasks::ConvertSchemasToTs => nil,
+        Test::Tasks::ConvertSchemasToTs => Test::Tasks::SetupDb,
```
davidrunger, line 38: (Unrelated change, but I happened to notice that (I think) `ConvertSchemasToTs` ultimately goes through Rails, which will error if the DB doesn't exist yet.)

lib/test/requirements_resolver.rb
```diff
@@ -33,22 +35,35 @@ def dependency_map
...
           Test::Tasks::CreateDbCopies,
           Test::Tasks::CompileJavaScript,
+          Test::Tasks::StartPercy,
+        ],
+        Test::Tasks::RunFeatureTestsB => [
+          Test::Tasks::DivideFeatureSpecs,
+          Test::Tasks::CreateDbCopies,
+          Test::Tasks::CompileJavaScript,
+          Test::Tasks::StartPercy,
+          Test::Tasks::RunApiControllerTests,
```
davidrunger, line 57: `RunFeatureTestsB` uses the `_api` database, so we have to wait until those tests are complete, so that there aren't conflicts over the DB with the API controller tests.

davidrunger:
Oh baby, the parallelism looks great and there appears to be quite a real speedup! This is a record-low build time!

https://github.com/davidrunger/david_runger/actions/runs/13281980014/job/37082138625?pr=6107

image

https://davidrunger.com/ci_step_results

image

image

That being said, both CodeCov and Percy failed, so I need to make sure that those issues can be resolved.

davidrunger:
Even better on the next run!

https://github.com/davidrunger/david_runger/actions/runs/13282307442/job/37083121181?pr=6107

image

https://davidrunger.com/ci_step_results

image

Percy still flaked, though: https://percy.io/David-Runger/web/david_runger/builds/38682378/changed/2093229736?browser=chrome&browser_ids=64&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_private&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

I guess that the browser is slower to render the SVGs, because there is more CPU contention, due to the increased concurrency?

PR #6114, opened 2025-02-12
[specs] Make CompileJavaScript await ConvertSchemasToTs

davidrunger:
Because the schema task [does `rm -rf` on a directory][1], there's a risk that it will delete a file that Vite needs to compile the JavaScript. This seems to have happened [here][2].

To avoid this issue, this change makes the `CompileJavaScript` task wait for the completion of the `ConvertSchemasToTs`.

`ConvertSchemasToTs` was pretty slow (10+ seconds), so this change would have had a pretty negative performance impact on CI build times. However, I think that the vast majority of its slowness was unnecessary and a result of going through a Rails runner bin script, rather than just invoking the relevant Ruby code directly. This change begins to invoke the script directly, which I think will save a huge amount of time, hopefully making this change not very costly to CI run times.

[1]: https://github.com/davidrunger/david_runger/blob/7ead899fdeba0749aec65d7129c1fca0e3023954/tools/json_schemas_to_typescript.rb/#L39 [2]: https://github.com/davidrunger/david_runger/actions/runs/13283567361/job/37087085374#step:11:120

### DIFF
--- a/lib/test/requirements_resolver.rb
+++ b/lib/test/requirements_resolver.rb
@@ -15,7 +15,10 @@ class Test::RequirementsResolver
         # Installation / Setup
         Test::Tasks::PnpmInstall => nil,
         Test::Tasks::BuildRouteHelpers => Test::Tasks::PnpmInstall,
-        Test::Tasks::CompileJavaScript => Test::Tasks::BuildRouteHelpers,
+        Test::Tasks::CompileJavaScript => [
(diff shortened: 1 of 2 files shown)


davidrunger:
It's crazy how much time we were wasting on `ConvertSchemasToTs`. This change brings it from ~10+ seconds to almost invisible on the Gantt charts:

image

PR #6120, opened 2025-02-13
[ci] Increase Pallets concurrency from 3 to 4 [DEV-175]

davidrunger:
This change increases the Pallets concurrency in CI from 3 to 4.

I think that this makes a certain amount of sense, because we have 4 vCPU cores.

This change is also is somewhat of a response to recent changes that broke up some steps into smaller sub-tasks, namely #6107 (feature specs) and especially #6118 (Vite builds). That's good, because it allows for more parallelization, but I think it can also cause some problems, because now more of our concurrency "slots" can be occupied by those now parallelized tasks, which can delay the execution of other tasks (e.g. DB setup) that are now waiting for a slot, whereas previously those other tasks (e.g. DB setup) might have been able to start executing sooner, because available concurrency slots were not being so occupied by parallelized tasks.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -79,7 +79,7 @@ jobs:
         run: bin/run-tests
         env:
           CI: true
-          PALLETS_CONCURRENCY: 3
+          PALLETS_CONCURRENCY: 4
           AWS_EC2_METADATA_DISABLED: true
           RAILS_ENV: test
           DISABLE_SPRING: 1
--- a/lib/test/run_pallets_tests.rb
+++ b/lib/test/run_pallets_tests.rb
@@ -4,7 +4,7 @@ require_relative 'middleware/task_result_tracking_middleware.rb'
 require_relative 'runner.rb'
 
 Pallets.configure do |c|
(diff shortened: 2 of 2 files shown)


davidrunger:
Here's a graph of the Gantt chart for this PR's first run:

image

I think that this change (increasing the Pallets concurrency from 3 to 4) did help significantly. However, there might still be some room for improvement?, although making these improvements might involve some subtle logic. RunPrettier appears to be blocking CreateDbCopies, and indeed for most of RunPrettier's execution time it is doing just that. However, RunPrettier was able to start sooner than CreateDbCopies. When RunPrettier started, BuildFixtures was not _quite_ yet done running, so CreateDbCopies couldn't yet run at that time. We could have delayed RunPrettier or something, and given that concurrency slot to CreateDbCopies instead. This probably would have resulted in an overall faster runtime. But trying to determine that ahead of time and encode that into the simple logic of our CI step running might not be easy.

For now, I guess I will just go ahead and merge this. Maybe I will continue looking for other ways to optimize our CI build times.

davidrunger:
With this change, for the build on `main`, we achieved a record-low 73.7 second build time!

image

image

PR #6148, opened 2025-02-14
Split feature tests into 3 groups (not 2)

davidrunger:
I think that this will better make use of the 4 vCPUs that are on GitHub Actions's machines. I tried this before (#6063), and it didn't seem to work, but I didn't have some things in place then that I have now (like more robust and efficient Percy setup for parallel runs, and also the Gantt charts, which are super helpful to understand what's going on), so I think it might work well this time around, i.e. I think it might decrease overall run time.

### DIFF
--- a/app/poros/redis_options.rb
+++ b/app/poros/redis_options.rb
@@ -28,7 +28,7 @@ class RedisOptions
       when '_unit', nil then 4
       when '_api' then 5
       when '_html' then 6
-      when '_feature_c' then 7
+      when '_feature' then 7
       else raise('Unexpected DB_SUFFIX!')
       end
     else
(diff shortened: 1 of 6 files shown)


davidrunger:
I think that this is a speed-up, at least on `main` (which, to some extent, is the only scenario that we can really compare against, because branch runs are variable in terms of which steps they run).

Since approximately 6am CT on Feb 12, we have these stats on `main` for the 28 builds since then:

```
== On Main
81.2 average
78.3 best sub-average
70.7 best run
```

I just ran this branch 3 times with the `main` set of steps, and got these times:

```
== On this branch
71.0
72.4
76.0
average: 73.1
```

That best run is very close to the overall best run from the most recent 28 builds on `main`. The average is over 5 seconds better than the best possible average on `main` that can be calculated from the present going back anywhere from 1 to 28 builds, and more than 8 seconds better than the most recent 28 builds on `main`. This suggests to me that we probably will see a slight performance gain on `main`. So, I will merge this.

PR #6151, opened 2025-02-15
[JsonSchemaValidator] Automatically open the schema file in editor

davidrunger:
This will make it a little easier to paste the JSON Schema into the appropriate file (since it will now be opened automatically in the developer's editor). We were already creating and opening the JSON Schema file if it didn't exist, but this change makes it so that we will also open the file even if it does already exist, for easier editing (or pasting in an updated version of the JSON Schema generated by the converter tool, which will also be opened automatically).

### DIFF
--- a/app/poros/json_schema_validator.rb
+++ b/app/poros/json_schema_validator.rb
@@ -20,7 +20,7 @@ class JsonSchemaValidator
         !(universal_bootstrap_data? && !schema_file_exists?) &&
         schema_validation_errors.present?
     )
-      copy_data_to_clipboard_and_open_schema_converter_if_development
+      facilitate_schema_provisioning_if_development
 
       raise(
         NonconformingData,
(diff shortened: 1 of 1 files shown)


app/poros/json_schema_validator.rb
```diff
@@ -118,6 +108,12 @@ def copy_data_to_clipboard_and_open_schema_converter_if_development
         puts('Copied JSON to clipboard.'.yellow)
       end
 
+      # Open schema file in editor (creating it, if needed).
+      FileUtils.mkdir_p(File.dirname(absolute_schema_path))
+      FileUtils.touch(absolute_schema_path)
+      system("$EDITOR '#{absolute_schema_path}'", exception: true)
```
davidrunger, line 110: I tried to open the editor before opening the converter in the browser, in the hope that the browser will then be "on top", but unfortunately VS Code is slow enough to open that it still ends up coming out "on top" at least some of the time. Not a huge deal, though, and I'm not sure there's much we can reasonably do to change this.

PR #6153, opened 2025-02-15
Speed up and otherwise improve feature specs [DEV-177]

davidrunger:
I think that the biggest performance improvement here comes from the new `sign_in_confirmed_via_my_account?` method.

`async_partial_delay_for_rails_env` also helps.

Changing some specs to `:rack_test_driver` helps.

Even changing `root_path` to `up_path` in `spec/features/rack_attack_spec.rb` helps.

Consolidating two specs in the workouts file helps a bit.

Some of these changes make our tests somewhat worse. For example, `async_partial_delay_for_rails_env` creates a divergence between the test environment and the real environment. Removing some `sleep`s that we had to try to catch problems that could theoretically emerge of a certain amount of time increases our risk of missing some bugs. But, on balance, I am estimating that these tradeoffs are overall worthwhile, in order to have faster feature specs.

Other changes are just cosmetic/formatting or otherwise minor changes.

### DIFF
--- a/spec/features/logs_spec.rb
+++ b/spec/features/logs_spec.rb
@@ -1,7 +1,7 @@
 RSpec.describe 'Logs app' do
   let(:user) { users(:user) }
 
-  context 'when user is signed in' do
+  context 'when a user is signed in' do
     before { sign_in(user) }
 
     let(:delete_last_entry_text) { 'Delete last entry' }
@@ -63,7 +63,7 @@ RSpec.describe 'Logs app' do
         end
         expect(page).to have_text('New Log')
 
-        # Check for (platform-specific) keyboard shortcut to open log-selector modal
+        # Check for (platform-specific) keyboard shortcut to open log-selector modal.
         expect(page).to have_text('Tip: Ctrl+K will open the log selector.')
 
         log = user.logs.first!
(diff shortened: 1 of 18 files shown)


davidrunger:
# Tip

View ignoring whitespace changes for a somewhat cleaner/simpler diff: https://github.com/davidrunger/david_runger/pull/6153/files?w=1

spec/features/logs_spec.rb
```diff
@@ -161,27 +161,25 @@
         end
       end
 
-      context 'when there is one entry for the log' do
-        before { log.log_entries.drop(1).each(&:destroy!) }
```
davidrunger, line 165: This `context` condition didn't really seem significant, so I deleted it.

spec/features/logs_spec.rb
```diff
@@ -279,7 +277,7 @@
           # negative (not failing). A wait time of 1 second seems to be long
           # enough to make the test consistently fail if this is broken.
           publish_new_log_entry
-          sleep(1)
+          sleep(0.5)
```
davidrunger, line 280: I'm leaving the comment mentioning that a 1 second sleep is reliable, but I'm decreasing this to 0.5 seconds to be faster but hopefully still be able to catch a bug at least intermittently if there is one.

spec/features/workout_spec.rb
```diff
@@ -4,28 +4,6 @@
   context 'when user is signed in' do
     before { sign_in(user) }
 
-    it 'renders expected content' do
```
davidrunger, line 7: I consolidated this example into the one below.

PR #6185, opened 2025-02-18
Save user's Google 'sub' value in DB and check it on login [DEV-122]

davidrunger:
This should help to protect users against domain takeover identity attacks as described here: https://cyberinsider.com/unfixed-google-oauth-flaw-exposes-millions-to-account-takeovers/ .

NOTE: That article claims the `sub` to be unreliable in 0.04% of cases. I am going to just hope that those cases won't include any of our users.

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -7,7 +7,21 @@ class Users::OmniauthCallbacksController < Devise::OmniauthCallbacksController
     email = access_token.info['email']
 
     user = User.find_by(email:)
-    if !user
+
+    if user
+      # https://cyberinsider.com/unfixed-google-oauth-flaw-exposes-millions-to-account-takeovers/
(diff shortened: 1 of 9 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -7,7 +9,29 @@ def google_oauth2
...
+              sub_in_google_response: sub,
+            },
+          )
+
+          flash[:alert] = 'You are attempting a domain identity takeover attack. Blocked!'
+          redirect_to(new_user_session_path)
+          return
+        end
+      else
+        user.update!(google_sub: sub)
```
davidrunger, line 32: We will gradually "backfill" users' Google `sub` values as they log in. Once a user logs in after this change has been deployed, then they should henceforth be protected from this identity takeover attack. Unfortunately, I don't think that there is a way to protect against this attack for users who don't log in after this change has been deployed.

PR #6197, opened 2025-02-20
[logs] Add 'dependent: :destroy' to LogEntry delegated_type

davidrunger:
Also, add a datamigration to destroy log entry datum records that were orphaned before adding this `dependent: :destroy` just now.

Also, add a generator to create timestamped datamigrations that will use a `Datamigration` library class (which includes a tagged logger and an easy mechanism to test and roll back the datamigration within a transaction for testing locally and repeatedly).

Also, while fiddling with custom cops config, apply `CustomCops/DontIncludeSidekiqWorker` only to `app/workers/**/*.rb` and exclude `tools/custom_cops/**/*.rb` from `Style/ClassAndModuleChildren`.

### DIFF
--- a/spec/tools/custom_cops/dont_include_sidekiq_worker_spec.rb
+++ b/spec/tools/custom_cops/dont_include_sidekiq_worker_spec.rb
@@ -2,19 +2,15 @@ require 'rubocop'
 require 'rubocop/rspec/support'
 require Rails.root.join('tools/custom_cops/dont_include_sidekiq_worker.rb')
 
-RSpec.describe CustomCops::DontIncludeSidekiqWorker do
+RSpec.describe CustomCops::DontIncludeSidekiqWorker, :config do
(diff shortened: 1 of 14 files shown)


spec/tools/custom_cops/dont_include_sidekiq_worker_spec.rb
```diff
@@ -2,19 +2,15 @@
 require 'rubocop/rspec/support'
 require Rails.root.join('tools/custom_cops/dont_include_sidekiq_worker.rb')
 
-RSpec.describe CustomCops::DontIncludeSidekiqWorker do
+RSpec.describe CustomCops::DontIncludeSidekiqWorker, :config do
```
davidrunger, line 5: (I simplified this spec slightly, because I was then going to use it as a basis for the new cop spec added in this PR, so I wanted to simplify it as much as reasonably possible, first.)

PR #6213, opened 2025-02-22
Provide a stable SECRET_KEY_BASE env var in development

davidrunger:
This will make it so that Docker rebuilds after file changes (and also, I think, deleting `tmp/`) will no longer log users out in local development.

https://chat.deepseek.com/a/chat/s/f54c613b-ca4e-42be-b7dc-4b2053d03600

https://x.com/i/grok?conversation=1893160209570210301

We move the file earlier in the initializers (using lexicographical ordering) because we want it to run before `config/initializers/devise.rb`, which will reference the `Rails.application.secret_key_base` value that seems to get set by `DavidRunger::Application.config.secret_key_base = [...]`. (This doesn't seem to strictly be required in order to fix the issue of users being logged out in local development, but having the Devise secret key not match the Rails application key, even though it would appear from the code in `config/initializers/devise.rb` that they should/will match, seems like a potential source of confusion/bugs that I would rather avoid.)

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -2,5 +2,6 @@ BROWSER=firefox-devedition # this helps letter_opener/launchy to open emails in 
 JWT_SECRET=a-dummy-development-jwt-secret
 RAILS_ENV=development
 RAILS_MAX_THREADS=1 # run web server without concurrency for more clearly structured log output
+SECRET_KEY_BASE=dummy-development-secret-key-base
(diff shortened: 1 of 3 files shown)


davidrunger:
This will be good because it will provide a more meaningful signal about whether or not any given change will actually log real users out in production. For example, when testing #6212, I noticed that my user seemed to be logged out locally, and I wasn't sure whether that was due to the change being made in that PR or some other factor. With this fix in place, it should be possible now to more confidently ship a change like that without worrying that it will log out users in production.

PR #6226, opened 2025-02-22
[ci] Create a dedicated database for RunFeatureTestsA

davidrunger:
This way, RunFeatureTestsA won't have to wait for RunUnitTests to complete before RunFeatureTestsA can start running. Having to wait used not to be a problem, because RunUnitTests always finished before CompileUserJavaScript, anyway, but that's not the case anymore.

The goal is to avoid runs like the one below, where RunFeatureTests A waits for RunUnitTests even though CompileUserJavaScript had already completed:

image

That run was particularly unlucky because RunFeatureTestsA happened to be the slowest of the three feature test runners, and it started last, due to waiting for the unit tests. With this change, there will no longer be any waiting for unit tests.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,3 +1,4 @@
 GOOGLE_OAUTH_CLIENT_ID=fake-google-oauth-client-id
 GOOGLE_OAUTH_CLIENT_SECRET=fake-google-oauth-client-secret
 RAILS_MAX_THREADS=2
+SECRET_KEY_BASE=dummy-test-secret-key-base
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -85,7 +85,6 @@ jobs:
           DISABLE_SPRING: 1
           PERCY_SKIP_UPDATE_CHECK: 1
           VITE_RUBY_AUTO_BUILD: false
-          SECRET_KEY_BASE: 0af8c0e49e9259f2d777cfcd121d8540cee914c34b0abf08dd825bdee47c402daf9f2b9b74a9fff6f88ed95257ae39504e0d59c66400413b0cf7d29079c6358b # gitleaks:allow This is just for the test environment; not a real secret.
           # The hostname used to communicate with the PostgreSQL service container
           POSTGRES_HOST: 127.0.0.1
           POSTGRES_USER: postgres
--- a/.runger-config.yml
+++ b/.runger-config.yml
@@ -4,5 +4,5 @@ spec-commands: |
   export RAILS_ENV=test PALLETS_CONCURRENCY=${PALLETS_CONCURRENCY:-5} DISABLE_SPRING=1
   export POSTGRES_USER=david_runger POSTGRES_HOST=localhost
   set -x
-  redis-cli -n 8 FLUSHDB
+  redis-cli -n 10 FLUSHDB
   bin/run-tests
--- a/app/poros/redis_options.rb
+++ b/app/poros/redis_options.rb
@@ -28,7 +28,8 @@ class RedisOptions
       when '_unit', nil then 4
       when '_api' then 5
       when '_html' then 6
-      when '_feature' then 7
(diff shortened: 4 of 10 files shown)


app/poros/redis_options.rb
```diff
@@ -28,7 +28,8 @@ def default_db_number
       when '_unit', nil then 4
       when '_api' then 5
       when '_html' then 6
-      when '_feature' then 7
+      when '_feature_a' then 7
+      when '_feature_c' then 9
```
davidrunger, line 32: I'm leaving `8` for if we ever add `_feature_b`.

.runger-config.yml
```diff
@@ -4,5 +4,5 @@ spec-commands: |
   export RAILS_ENV=test PALLETS_CONCURRENCY=${PALLETS_CONCURRENCY:-5} DISABLE_SPRING=1
   export POSTGRES_USER=david_runger POSTGRES_HOST=localhost
   set -x
-  redis-cli -n 8 FLUSHDB
+  redis-cli -n 10 FLUSHDB
```
davidrunger, line 7: Increasing to 10 so as not to conflict with the DBs in `app/poros/redis_options.rb`. We aren't actually using `8` yet, because RunFeatureTestsB is sharing the `_api` DB, but we might in the future.

lib/test/run_pallets_tests.rb
```diff
@@ -9,7 +9,7 @@
 
   c.concurrency = concurrency
   c.max_failures = 0
-  c.backend_args = { url: 'redis://127.0.0.1:6379/8' } # use redis db #8 (to avoid conflicts)
+  c.backend_args = { url: 'redis://127.0.0.1:6379/10' } # use redis db #10 (to avoid conflicts)
```
davidrunger, line 12: See https://github.com/davidrunger/david_runger/pull/6226/files#r1966494039 .

lib/test/tasks/create_db_copies.rb
```diff
@@ -5,8 +5,10 @@ def run
     # The commands below will error if there are any active connections to the
     # database, so disconnect.
     ActiveRecord::Base.connection_pool.connections.each(&:disconnect!)
+    # Give a little time for the disconnections to complete (?).
+    sleep(0.5)
```
davidrunger, line 9: See this commit message for an explanation of this addition: https://github.com/davidrunger/david_runger/pull/6226/commits/fbbc66f2b9b9807e1107ecdfcacc5bf6e8931243 (which I think might actually not be really related to this change?, even though it seems super suspicious that I was already changing this file on the lines right below here).

.env.test
```diff
@@ -1,3 +1,4 @@
 GOOGLE_OAUTH_CLIENT_ID=fake-google-oauth-client-id
 GOOGLE_OAUTH_CLIENT_SECRET=fake-google-oauth-client-secret
 RAILS_MAX_THREADS=2
+SECRET_KEY_BASE=dummy-test-secret-key-base
```
davidrunger, line 4: Explained here: https://github.com/davidrunger/david_runger/pull/6226/commits/b97d88fcce1361c9ba4fcd2c960aa0eb40f3b0e7

.github/workflows/ruby.yml
```diff
@@ -85,7 +85,6 @@ jobs:
           DISABLE_SPRING: 1
           PERCY_SKIP_UPDATE_CHECK: 1
           VITE_RUBY_AUTO_BUILD: false
-          SECRET_KEY_BASE: 0af8c0e49e9259f2d777cfcd121d8540cee914c34b0abf08dd825bdee47c402daf9f2b9b74a9fff6f88ed95257ae39504e0d59c66400413b0cf7d29079c6358b # gitleaks:allow This is just for the test environment; not a real secret.
```
davidrunger, line 88: See https://github.com/davidrunger/david_runger/pull/6226/files#r1966495665

PR #6228, opened 2025-02-22
Increase Pallets concurrency from 5 to 6

davidrunger:
There are some times when we are bound by the concurrency limit of 5 and would to better to be able to run 6 tasks simultaneously. This is particularly due to #6223, which causes the `StartPercy` task to take much longer. Since I think that task largely uses IO (waiting on network calls to the Percy server and/or sleeping), it's probably not using the CPU very much when it runs, so I think that (at least during that time) adding another concurrent task won't add much to the CPU burden.

This change increases the `sleep` in `StartPercy` iterations from 2 seconds to 4 seconds to try to make the CPU burden of `StartPercy` even lower. The only thing that really matters is that it completes before `CompileUserJavaScript` finishes, which I think that it still will do.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -79,7 +79,7 @@ jobs:
         run: bin/run-tests
         env:
           CI: true
-          PALLETS_CONCURRENCY: 5
+          PALLETS_CONCURRENCY: 6
           AWS_EC2_METADATA_DISABLED: true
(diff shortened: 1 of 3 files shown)


davidrunger:
The goal of this PR is that, in a build like this, we don't want `RunUnitTests` to have to wait for `StartPercy`:

image

PR #6232, opened 2025-02-22
[home#index monitor] Tighten median expectation to 5..50 ms

davidrunger:
I received an email today saying that our responses are too fast!

> The DataMonitors::HomeIndexRequests#median_response_time_in_past_day check failed to match its expected value.
> Actual value: 9.0
> Expected value: 10..100
> Checked at: 2025-02-22 07:07:12 -0600 

This is because our new Hetzner server is speedy!

This change updates the expectation to be more in line with our fast new server, lowering the expected median response time from a range of `10..100` to `5..50`.


davidrunger:
We switched over to Hetzner on the morning of Feb 21:

image

PR #6235, opened 2025-02-23
Add endpoint to forward Grafana alert webhooks as emails [DEV-186]

davidrunger:
I want to make it so that Grafana can send a webhook to my app when a monitoring alert is triggered, and my app will convert that webhook content into an email and send it to me. (There's something unideal about using the system that is being monitored as part of the alert-sending process, but I'll take that risk.)

I'm doing this instead of just having Grafana send the email because it seems that configuring Grafana to send emails itself is a decently big pain and/or somewhat of a security risk (e.g. having to put a GMail password for my personal account onto the server).

To develop and test the Grafana -> `web` webhook, I wanted to be able to iterate quickly on changes to the Rails web app while running the Rails web app within the Docker Compose context (where it could be pinged by Grafana's alert test requests). I didn't want to have to rebuild the Docker image each time that I made a change to the Rails app source code, though; that would have been very slow and annoying!

Fortunately, Docker Compose has something that seems to be designed for this use case, [Compose Watch][1].

[1]: https://docs.docker.com/compose/how-tos/file-watch/

We can use Docker Compose Watch via a command like this...

```
COMPOSE_BAKE=true GIT_REV=$(git rev-parse origin/main) RUBY_VERSION=$(cat .ruby-version) RAILS_ENV=development docker compose watch cadvisor clock grafana loki nginx node_exporter postgres prometheus rails_metrics redis-app redis-cache vector web worker
```

... which will boot the specified services (which is all that run in production, except for `certbot`, which there's not really a point in running locally) and which will sync file changes to the running Docker images that were built with a `develop.watch` configuration in `docker-compose.yml`, which is currently the services that use the `default-rails-config`.

We cannot use `--build-arg` with Docker Compose Watch, though, so this change switches from using `--build-arg` for the `docker compose build` commands in `bin/build-docker` to instead pass environment variables which (per the changes herein to `docker-compose.yml`) will be pulled in via `build.args`.

Also, so that Rails won't reject the requests made within the Docker Compose network to `http://web:3000` (in either development or production), this change also adds `web:3000` as a permitted host in `config/application.rb`.

This change also expands the `TokenAuthenticatable` concern to look for a possible auth token not only in `params`, as we were doing before, but also in an `Authorization: Bearer <the secret here>` header, since this is one of Grafana's two options for sending an authentication secret (the other being basic auth).

Other than that, this change basically just adds a simple route, controller action, policy, and mailer to forward the webhook request content as an email.

### DIFF
--- a/app/policies/webhook_email_forward_policy.rb
+++ b/app/policies/webhook_email_forward_policy.rb
@@ -1,0 +1,2 @@
+class WebhookEmailForwardPolicy < ApplicationPolicy
+end
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,10 +7,19 @@ x-logging-config: &default-logging
 
 x-rails-config: &default-rails-config
   build:
+    args:
+      - GIT_REV=${GIT_REV}
+      - RUBY_VERSION=${RUBY_VERSION}
+      - RAILS_ENV=${RAILS_ENV}
     context: .
   depends_on:
     initialize_database:
       condition: service_completed_successfully
+  develop:
+    watch:
+      - action: sync
+        path: .
+        target: /app
   env_file:
     # NOTE: We need this to have DATABASE_URL in bin/docker-entrypoint.
     - .env.production.local
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,8 +9,8 @@ x-rails-config: &default-rails-config
   build:
     args:
       - GIT_REV=${GIT_REV}
-      - RUBY_VERSION=${RUBY_VERSION}
       - RAILS_ENV=${RAILS_ENV}
+      - RUBY_VERSION=${RUBY_VERSION}
     context: .
   depends_on:
     initialize_database:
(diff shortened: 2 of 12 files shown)


app/policies/webhook_email_forward_policy.rb
```diff
@@ -0,0 +1,2 @@
+class WebhookEmailForwardPolicy < ApplicationPolicy
```
davidrunger, line 1: (NOTE: Most of our policies refer to an ActiveRecord model; this one does not.)

docker-compose.yml
```diff
@@ -7,10 +7,19 @@ x-logging-config: &default-logging
...
+      - RUBY_VERSION=${RUBY_VERSION}
     context: .
   depends_on:
     initialize_database:
       condition: service_completed_successfully
+  develop:
+    watch:
+      - action: sync
+        path: .
+        target: /app
```
davidrunger, line 22: There's actually an official Docker article about using Docker with Ruby on Rails: https://docs.docker.com/guides/ruby/develop/

That article suggests `action: rebuild` rather than `action: sync`. I'm not totally sure why. From reading the docs, it seems like `sync` will be much faster. It seems to get the job done well for me.

PR #6246, opened 2025-02-24
[ci] Run unit tests only after API controller tests have completed

davidrunger:
My thinking is that this will hopefully reduce the parallel CPU demands during JavaScript compilation, allowing feature tests to start sooner. It's not clear whether the implications of this will be an overall win for test suite run time, but I think it might be, and I think it might be worth giving this a shot.

PR #6245, opened 2025-02-24
[ci] Set PALLETS_CONCURRENCY to 1

davidrunger:
I want to see what the task runtimes are like without any concurrency.

PR #6247, opened 2025-02-24
[ci] Run API controller tests after StartPercy completes

davidrunger:
This change builds on #6246 to further try to reduce concurrency while JavaScript assets are being compiled. My hope is that this will make JavaScript compilation faster, allowing feature tests to start sooner, and that this might be worth the price paid in terms of higher concurrency during feature specs and a later start for unit tests and API controller tests. It's not clear whether the implications of this will be an overall win for test suite run time, but I think it might be, and I think it might be worth giving this a shot.

My inspiration for this change is largely #6245, which showed that, when PALLETS_CONCURRENCY is 1, we can have these run times:

```
CompileAdminJavaScript : exited with 0 (took 12.866 seconds)
 CompileUserJavaScript : exited with 0 (took 16.567 seconds)
```

That's only 29.4 seconds total, which is faster than even just the `CompileUserJavaScript` steps takes alone when running with PALLETS_CONCURRENCY of 6. So, the clear implication seems to be that lower concurrency during JavaScript builds can shorten the build time significantly. This change aims to accomplish that (lower concurrency during JavaScript builds).

### DIFF
--- a/lib/test/requirements_resolver.rb
+++ b/lib/test/requirements_resolver.rb
@@ -53,7 +53,12 @@ class Test::RequirementsResolver
           Test::Tasks::CreateDbCopies,
           Test::Tasks::RunApiControllerTests,
         ],
-        Test::Tasks::RunApiControllerTests => Test::Tasks::CreateDbCopies,
+        # RunApiControllerTests doesn't really depend on StartPercy,
+        # but it's better for the timing of steps if it waits for it.
+        Test::Tasks::RunApiControllerTests => [
+          Test::Tasks::CreateDbCopies,
+          Test::Tasks::StartPercy,
+        ],
         Test::Tasks::RunFileSizeChecks => Test::Tasks::CompileUserJavaScript,
         Test::Tasks::RunFeatureTestsA => [
           Test::Tasks::DivideFeatureSpecs,


davidrunger:
Results on this branch:

image

```
      UploadViteAssets : exited with 0 (took  0.000 seconds)
            RunRubocop : exited with 0 (took  0.002 seconds)
     RunFileSizeChecks : exited with 0 (took  0.004 seconds)
    DivideFeatureSpecs : exited with 0 (took  0.011 seconds)
    ConvertSchemasToTs : exited with 0 (took  0.083 seconds)
  FeatureTestsCanStart : exited with 0 (took  0.101 seconds)
          RunTypelizer : exited with 0 (took  0.222 seconds)
         CheckVersions : exited with 0 (took  0.508 seconds)
         BuildFixtures : exited with 0 (took  1.045 seconds)
             StopPercy : exited with 0 (took  1.142 seconds)
        CreateDbCopies : exited with 0 (took  1.281 seconds)
     BuildRouteHelpers : exited with 0 (took  1.432 seconds)
               SetupDb : exited with 0 (took  3.611 seconds)
           PnpmInstall : exited with 0 (took  3.825 seconds)
           RunBrakeman : exited with 0 (took  8.258 seconds)
 RunApiControllerTests : exited with 0 (took 12.253 seconds)
RunHtmlControllerTests : exited with 0 (took 17.888 seconds)
            StartPercy : exited with 0 (took 18.252 seconds)
CompileAdminJavaScript : exited with 0 (took 18.594 seconds)
 CompileUserJavaScript : exited with 0 (took 29.434 seconds)
          RunUnitTests : exited with 0 (took 34.966 seconds)
      RunFeatureTestsB : exited with 0 (took 36.304 seconds)
      RunFeatureTestsC : exited with 0 (took 38.836 seconds)
      RunFeatureTestsA : exited with 0 (took 40.133 seconds)

Wall clock elapsed time: 76.006
Total task time: 268.184
Parallelism: 3.528
```

There's a lot more concurrency now when feature specs are running, which presumably slows down those steps somewhat. I'm not sure it's worth it. But let's merge this and see what happens.

PR #6249, opened 2025-02-24
Use spring in test

davidrunger:


davidrunger:
It seems like this does make things faster, but it also seems to make things really flaky in a lot of different ways. And the code that I've implemented so far to try to fight against this flakiness (mostly my `Spring.after_fork` block) seems sort of brittle and hacky. I don't think it's really worthwhile to try to keep going down this road.

PR #6262, opened 2025-02-24
Move from Rails credentials to env vars in production [DEV-184]

davidrunger:
I have not had a very good experience with Rails credentials, and I think it will be better just to manage credentials via env vars. There are already some other secrets that we are managing via environment variables, so I think that we might as well just manage all production secrets via environment variables.

For now, though, we are still using Rails credentials in `test`, because I currently cannot decrypt those values, and also I think that there's a better argument for using Rails credentials in test, because it's not so easy to provide environment variables in test as it is in `development` and `production`.

I have already added the env vars referenced in this change to `.env.production.local` on the server.

After this deploys, I will delete the `RAILS_MASTER_KEY` env var from `.env.production.local` that we are currently using to decrypt the credentials in production.


.env.production
```diff
@@ -1 +1,3 @@
 RAILS_MAX_THREADS=2
+# https://island94.org/2025/01/ruby-thread-contention-simply-gvl-queuing
+RUBY_THREAD_TIMESLICE=10
```
davidrunger, line 3: (Mostly unrelated, but I noticed that this was in `.env.production.local` with my secret env vars, but there's no reason for it to be secret, so I'm moving it here.)

PR #6263, opened 2025-02-24
Don't require SECRET_KEY_BASE if SECRET_KEY_BASE_DUMMY=1

davidrunger:
This should fix deploys, which are currently failing. https://github.com/davidrunger/david_runger/actions/runs/13503267388/job/37726931183#step:5:96

> KeyError: key not found: "SECRET_KEY_BASE" (KeyError)


davidrunger:
This is a followup to #6262.

PR #6278, opened 2025-02-25
Add user agent to browser-originating events [DEV-185]

davidrunger:
Also, add the pretty user agent to the admin index and show pages (and the raw agent to the show page).

I'm allowing the field to be nullable because I don't want all Events to necessarily originate from a browser.

### DIFF
--- a/app/decorators/request_decorator.rb
+++ b/app/decorators/request_decorator.rb
@@ -1,5 +1,4 @@
 class RequestDecorator < Draper::Decorator
-  prepend Memoization
   include UserAgentDecoratable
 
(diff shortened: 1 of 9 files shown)


app/decorators/request_decorator.rb
```diff
@@ -1,5 +1,4 @@
 class RequestDecorator < Draper::Decorator
-  prepend Memoization
```
davidrunger, line 2: (Unrelated change. I copied EventDecorator from this RequestDecorator example. While here, I noticed that we are including Memoization unnecessarily.)

PR #6292, opened 2025-02-27
Add ability to comment on blog posts [COM-1]

davidrunger:
This ended up taking a lot more than I expected. The comments work was a bit more involved than I expected, and also some other changes scope-creeped into this.

One of the biggest scope-creep changes was some changes to the `Intersection` type. These changes make that type _far_ more complex than it was before. I haven't even reviewed that in detail; ChatGPT wrote it for me: https://chatgpt.com/c/67c03a03-cc4c-8007-9962-6c22ed3bb313 . But the upshot is that I think it works better than it did before, in a few ways. For one thing, it now works for array input types, whereas before it basically didn't meaningfully work for them (apart from effectively making sure that both elements are arrays, I think).

Another notable scope-creep change is the addition of `decorate_with` functionality to our serializers. Unfortunately, this involved some light monkeypatching of Alba. But I think it will be good to be able to use decorators in serializers. I'm a little surprised that I haven't introduced this already.

I also added a good amount of logic related to redirecting users through a wizard flow after they log in. This code is kind of ugly/hacky, and could probably do well to be extracted to a concern and/or to be abstracted into some class(es) or something. For now, though, it serves the purpose (giving new users the opportunity to choose a display name), and so I am just going to go with it.


app/controllers/my_account_controller.rb
```diff
@@ -5,6 +5,7 @@ def destroy
     @user =
       User.includes(
         logs: %i[log_entries log_shares],
+        quiz_participations: %i[quiz_question_answer_selections],
```
davidrunger, line 8: (Basically an unrelated change to fix a Bullet complaint that was causing an exception when I tried to delete my account in order to test how that would work re: comments made.)

app/javascript/logs/components/EditLogSharingSettingsModal.vue
```diff
@@ -70,21 +70,28 @@ const logSharesSortedByLowercasedEmail = computed((): Array<LogShare> => {
     .sort((a, b) => a.email.toLowerCase().localeCompare(b.email.toLowerCase()));
 });
 
-const shareableUrl = computed((): string => {
-  return (
-    window.location.origin +
-    user_shared_log_path((bootstrap as Bootstrap).current_user.id, log.slug)
-  );
+const shareableUrl = computed((): string | null => {
```
davidrunger, line 73: (Mostly unrelated changes downstream from changes made to `Intersection`.)

app/javascript/types/bootstrap/LogsIndexBootstrap.ts
```diff
@@ -122,7 +122,9 @@ const schema = {
         },
         "log_shares": {
           "type": "array",
-          "items": {}
+          "items": {
```
davidrunger, line 125: (More mostly unrelated changes downstream from changes made to `Intersection`.)

app/controllers/api/comments_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def destroy
+    comment = current_user.comments.find(params[:id])
+
+    authorize(comment)
+
+    if comment.children.exists?
+      comment.update!(content: '[deleted]', user: nil)
+      render json: comment
```
davidrunger, line 26: I am not using `render_schema_json` here because I think it would be too difficult/annoying, with the fact that we sometimes respond with comment JSON but sometimes with `head :no_content`.

PR #6293, opened 2025-02-28
Update npm packages

davidrunger:
`pnpm update`

Dependabot doesn't (yet) support pnpm version 10, which we are using, so I am updating packages manually for now.


lib/test/tasks/run_file_size_checks.rb
```diff
@@ -24,7 +24,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'logs*.css' => (90..100),
     'logs*.js' => (793..803),
     'quizzes*.js' => (123..133),
-    'styles*.css' => (26..36),
+    'styles*.css' => (14..24),
```
davidrunger, line 27: I think that the file size decrease is probably due to this: https://github.com/tailwindlabs/tailwindcss/blob/main/CHANGELOG.md#408---2025-02-21

> Don't include theme variables that aren't used in compiled CSS (`https://github.com/tailwindlabs/tailwindcss/pull/16211`, `https://github.com/tailwindlabs/tailwindcss/pull/ 16676`)

Thanks! 🎉

PR #6294, opened 2025-02-28
Add @prettier/plugin-pug

davidrunger:
https://github.com/prettier/plugin-pug

Prettier is great for TypeScript. Let's use it for pug (in our Vue templates), as well.

I wasn't very happy with the formatting out of the box, but with these settings it does pretty much what I want:

```
  "pugSingleQuote": false,
  "pugAttributeSeparator": "none",
  "pugWrapAttributesThreshold": 1,
```

I will probably add this commit to `.git-blame-ignore-revs` after merging.

The biggest changes are that attributes are now wrapped in double quotes, which is a change that I have sort of been wanting to make for a while now. Reasons:

1. For JavaScript, we use single quotes. By wrapping pug attributes in double quotes, then when we write JavaScript code in the attribute (`v-if="checkAttribute('something')"`) we can stay consistent in using single quotes in the JavaScript.
2. I think that when browsers render HTML attributes they do so with double quotes, and so it feels sort of good to align with that in the source code.

The other notable changes are from `"pugWrapAttributesThreshold": 1` (which means that, if there is one attribute, then the attributes will be inlined, but if there is more than one attribute, then each will be broken onto a separate line), which I have somewhat been trying to follow, but which I haven't followed consistently.

I have already been pretty consistently following `"pugAttributeSeparator": "none"`, so that doesn't really change anything, relative to my existing code (though it is a significant change versus the `@prettier/plugin-pug` default, which adds commas between attributes).


app/javascript/groceries/components/Sidebar.vue
```diff
@@ -1,46 +1,46 @@
...
+              v-for="store in groceriesStore.sortedSpouseStores"
+              :store="store"
             )
     .mt-auto.text-center.p-3(
-      v-if='!groceriesStore.sortedSpouseStores.length && !collapsed'
+      v-if="!groceriesStore.sortedSpouseStores.length && !collapsed"
     )
       | Tip: You and your partner can automatically view each other's lists.
-      | #[a(:href='new_marriage_path()') Click here] to invite them to join.
+      |         #[a(:href="new_marriage_path()") Click here] to invite them to join.
```
davidrunger, line 43: Not sure why it's adding this indentation? 🤷‍♂️ 🤔

PR #6298, opened 2025-03-01
Make RunTypelizer depend on CreateDbCopies

davidrunger:
I am hoping that this will avoid errors like this:

```
createdb: error: database creation failed: ERROR:  source database "david_runger_test" is being accessed by other users
DETAIL:  There are 2 other sessions using the database.
'createdb -T david_runger_test david_runger_test_feature_a -U postgres -h 127.0.0.1 --no-password' with ENV vars {} failed (exited with 1, took 5.087).
```

-- https://github.com/davidrunger/david_runger/actions/runs/13604936208/job/38035323653?pr=6296#step:12:202


davidrunger:
Didn't work, at least not fully. Trying this: https://github.com/davidrunger/david_runger/pull/6299 .

I am going to leave this in place, though, because I think it might still be a good idea, even if it doesn't completely solve the problem. It might partially solve the problem. Not sure.

PR #6300, opened 2025-03-01
[CreateDbCopies] Add repeated attempts to disconnect & debug logging

davidrunger:


davidrunger:
I'm hoping that this will resolve (or at least help to debug) the issue that #6298 and #6299 were trying to fix. Also, hopefully most of the time this will keep the sleep at 0.1 seconds rather than 0.5 seconds.

PR #6309, opened 2025-03-02
Remove lint rule to test #6308

davidrunger:


davidrunger:
It works:

image

Closing.

PR #6314, opened 2025-03-03
Use quicktype to commit TypeScript types for JSON schemas [DEV-196]

davidrunger:
## Pros

1. We don't have to commit verbose JSON Schemas twice (once in `spec/support/schemas/` and then again copied into `app/javascript/types/`). Instead, the TypeScript types created by `quicktype` are much more succinct.
2. The TypeScript types are also easier for me to read.
3. Related to 1 and 2 above, I think we are less likely to have bugs when using the TypeScript types. Switching to TypeScript types helped me to see [this bug](https://github.com/davidrunger/david_runger/pull/6314/commits/9183a80645953165418fc5e6e8e893062c2f09df#diff-0101883617b7c2d0ae30fcfffae9daa6a7a66351a9e002b6d59d8db470d28b70L18) that had been unnoticed heretofore.
4. `quicktype` could maybe be used as a replacement for https://jsonformatter.org/json-to-jsonschema , and we could more so automate the whole flow rather than copy-pasting into and out of that website.

## Cons

1. Using `quicktype` is significantly slower than just copying the JSON Schema into a TypeScript type that parses it using `json-schema-to-ts`. I have tried to compensate for this by only generating the TypeScript type file that corresponds to a changed or added JSON Schema file, and by not generating the TypeScript types when booting the Rails server, and only running the conversion in CI when needed. This might miss some cases where compiling the types would be needed/useful, though.
2. `quicktype` is a much heavier dependency than `json-schema-to-ts` which it replaces.
3. `quicktype` does not seem like a very polished or well maintained library. There are warnings about `punycode` being deprecated ( https://github.com/glideapps/quicktype/issues/ 2662 ). It doesn't handle arrays properly ( https://github.com/glideapps/quicktype/issues/ 2481 ); I hacked a fix into this PR. It treats strings that have a `date-time` format as `Date`s ( https://github.com/glideapps/quicktype/pull/ 1151#issuecomment-602743429 ); I hacked a fix into this PR.

I guess let's try it. If it seems overall negative, we can revert. I'm hoping that it will be a positive, though.

---

This PR expands on the logic of the `Intersection` type. The reason for this is that, as mentioned above, `quicktype` (wrongly, IMO) converts JSON Schema `date-time` strings to `Date` types. TypeScript complained about some of the resulting types, but others it did not complain about, even though it seemed to me that TypeScript should have complained about them, since basically we were trying to intersect a type like `{ created_at: string }` with `{ created_at: Date }`. This change makes it so that such an intersection will be complained about.


lib/test/tasks/create_db_copies.rb
```diff
@@ -32,7 +35,7 @@ def run
       end
 
       execute_system_command(<<~COMMAND.squish)
-        psql -U postgres -h 127.0.0.1 -d david_runger_test -c "
+        psql -U #{postgres_user} -h #{postgres_host} -d david_runger_test -c "
```
davidrunger, line 38: This should have been done in #6306 . Since that PR, I believe that this step has been failing locally. I didn't run this step locally until now.

lib/test/tasks/run_html_controller_tests.rb
```diff
@@ -9,6 +9,7 @@ def run
       $(ls -d spec/controllers/*/ | grep -v 'spec/controllers/api/' | tr '\n' ' ')
       $(ls spec/controllers/*.rb)
       $(ls spec/helpers/*.rb)
+      $(ls spec/requests/{**/*.rb,*.rb})
```
davidrunger: See 598ebb1f1f4e704133d4aabf420e39a9d0db0449 .

lib/test/tasks/run_unit_tests.rb
```diff
@@ -2,14 +2,15 @@ class Test::Tasks::RunUnitTests < Pallets::Task
...
+    # Run all tests in `spec/` _except_ those in
+    # `spec/{controllers,helpers,requests}/` (which will we run by
+    # RunApiControllerTests and RunHtmlControllerTests) and `spec/features/`
+    # (which will be run by RunFeatureTests).
     execute_rspec_command(<<~'COMMAND')
       DB_SUFFIX=_unit
       bin/rspec
       $(ls -d spec/*/ |
-        grep --extended-regex -v 'spec/(controllers|features|helpers)(/|$)' |
+        grep --extended-regex -v 'spec/(controllers|features|helpers|requests)(/|$)' |
```
davidrunger, line 13: See 598ebb1f1f4e704133d4aabf420e39a9d0db0449 .

spec/requests/blog_spec.rb
```diff
@@ -2,7 +2,7 @@
   include BlogSpecHelpers
 
   describe 'GET /blog/some-article' do
-    subject(:get_blog_feed) { get('/blog/some-article', headers:) }
+    subject(:get_blog_article) { get('/blog/some-article', headers:) }
```
davidrunger, line 5: (Mostly unrelated, but I was looking at this spec since it started flaking in this PR. This subject was named incorrectly/badly/misleadingly (due to a lazy copy-paste, I think).)

spec/requests/blog_spec.rb
```diff
@@ -13,11 +13,11 @@
         end
       end
 
-      context 'when there is an Accept header for HTML, and Atom formats' do
+      context 'when there is an Accept header for HTML and Atom formats' do
```
davidrunger, line 16: (Mostly unrelated, but I was looking at this spec since it started flaking in this PR. The removed comma should not have been here.)

spec/support/schemas/bootstrap/logs/index.json
```diff
@@ -15,7 +15,7 @@
         },
         "toast_messages": {
           "type": "array",
-          "item": {
+          "items": {
```
davidrunger, line 18: This was an error in the JSON Schema. Looking at the TypeScript types made this error jump out. This is one of the benefits of committing the TypeScript types rather than a copy of the JSON Schema.

tools/json_schemas_to_typescript.rb
```diff
@@ -21,64 +21,116 @@ def initialize_listener(app)
...
+              removed: removed.map { relative_path(it) },
+            )
           end
 
         listener.start
       end
-
-      app.config.to_prepare do
-        JsonSchemasToTypescript.write_files
-      end
```
davidrunger, line 33: Removing type compilation on Rails server startup because the type compilation is so slow.

tools/json_schemas_to_typescript.rb
```diff
@@ -21,64 +21,116 @@ def initialize_listener(app)
...
       app.config.after_initialize do
         listener =
-          ::Listen.to(Rails.root.join(SCHEMA_DIRECTORY).to_s) do |_changed, _added, _removed|
-            JsonSchemasToTypescript.write_files
+          ::Listen.to(Rails.root.join(SCHEMA_DIRECTORY).to_s) do |changed, added, removed|
+            JsonSchemasToTypescript.write_files(
+              changed: changed.map { relative_path(it) },
+              added: added.map { relative_path(it) },
+              removed: removed.map { relative_path(it) },
+            )
```
davidrunger, line 29: Type compilation is much slower now, so rather than just always generating all types, we now want to be selective, and only generate the necessary types, depending on what has changed.

tools/json_schemas_to_typescript.rb
```diff
@@ -21,64 +21,116 @@ def initialize_listener(app)
...
+                --just-types
+            SH
+
+            # Only export the first interface in the file. (All are exported by quicktype.)
+            # Swap `Date` to `string`.
+            content_to_write =
+              File.read(file_to_write).
+                gsub(/^export interface/, 'interface').
+                sub(/^interface/, 'export interface').
+                gsub(/\bDate\b/, 'string')
```
davidrunger, line 87: Here we work around one bug in `quicktype`.

tools/json_schemas_to_typescript.rb
```diff
@@ -21,64 +21,116 @@ def initialize_listener(app)
...
+              CONTENT
+
+            File.write(
+              file_to_write,
+              content_to_write,
+            )
+
+            # Work around this quicktype bug:
+            # https://github.com/glideapps/quicktype/issues/ 2481
+            if JSON.parse(File.read(schema_path))['type'] == 'array'
```
davidrunger, line 103: Here we work around another bug in `quicktype`.

tools/json_schemas_to_typescript.rb
```diff
@@ -21,64 +21,116 @@ def initialize_listener(app)
...
       end
     end
-    # rubocop:enable Metrics/MethodLength
+    # rubocop:enable Metrics
+
+    private
+
+    def relative_path(absolute_path)
+      Pathname.new(absolute_path).relative_path_from(Rails.root).to_s
+    end
```
davidrunger, line 134: The DRY zealot in me wants to extract this to some shared location. We do basically the same thing here: https://github.com/davidrunger/david_runger/blob/7b24cbeb0156486c4ae84d3cace58bf1bd7bbbcc/lib/generators/datamigration/datamigration_generator.rb/#L17 .

PR #6322, opened 2025-03-04
Move back to upstream typelizer and bump from 0.1.0 to 0.3.0

davidrunger:


davidrunger:
Doesn't seem to work. Not 100% sure what the problem is. I think that maybe the app is being eager-loaded while Typelizer is disabled (when we create the DB), which causes problems.

davidrunger:
Going with https://github.com/davidrunger/typelizer/pull/15 and https://github.com/davidrunger/david_runger/pull/6325 , instead.

PR #6324, opened 2025-03-05
[SetupDb] Remove 'RAILS_ENV=test' arg that does nothing

davidrunger:
(... as far as I can tell.)

### DIFF
--- a/lib/test/tasks/setup_db.rb
+++ b/lib/test/tasks/setup_db.rb
@@ -3,7 +3,7 @@ class Test::Tasks::SetupDb < Pallets::Task
 
   def run
     execute_rake_task('db:create')
-    execute_rake_task('db:environment:set', 'RAILS_ENV=test')
+    execute_rake_task('db:environment:set')
     execute_rake_task('db:schema:load')
   end
 end


lib/test/tasks/setup_db.rb
```diff
@@ -3,7 +3,7 @@ class Test::Tasks::SetupDb < Pallets::Task
 
   def run
     execute_rake_task('db:create')
-    execute_rake_task('db:environment:set', 'RAILS_ENV=test')
+    execute_rake_task('db:environment:set')
```
davidrunger, line 6: For one thing, we already have `RAILS_ENV=test` in the environment from the GitHub Actions workflow: https://github.com/davidrunger/david_runger/blob/719fd6cac5a1de91faf77c11884116b8ff3e290f/.github/workflows/ruby.yml#L94

For another thing, this is an argument for the rake task, not an environment variable https://github.com/davidrunger/david_runger/blob/719fd6cac5a1de91faf77c11884116b8ff3e290f/lib/test/task_helpers.rb#L96 https://github.com/davidrunger/david_runger/blob/719fd6cac5a1de91faf77c11884116b8ff3e290f/lib/test/task_helpers.rb#L108 and I'm pretty sure that, as an argument, it does nothing.

PR #6332, opened 2025-03-05
[RunRubocop] Use tap formatter

davidrunger:
RuboCop has been really slow since re-enabling it:

image

I'm hoping that this tap formatter (hopefully just temporarily) might give some insight into why that is.


davidrunger:
I think I found a fix here: https://github.com/davidrunger/david_runger/pull/6334 .

PR #6333, opened 2025-03-05
[RunRubocop] Print paths that we are excluding

davidrunger:
I'm slightly suspicious that sometimes this takes a pathological form of some sort?


davidrunger:
This was for debugging. It accomplished its purpose, and resulted in https://github.com/davidrunger/david_runger/pull/6334. Closing.

PR #6337, opened 2025-03-05
[ci] Extract 'tools/' tests to standalone task

davidrunger:
Requiring the rubocop rspec stuff seems to immensely slow down the whole test suite. Let's isolate that and only run those tests when needed.


davidrunger:
Thanks to [vernier](https://github.com/jhawthorn/vernier), [profile-viewer](https://github.com/tenderlove/profiler/tree/ruby), and [test-prof](https://test-prof.evilmartians.io/) for helping me to see that RuboCop was slowing things down a lot!

I'm not sure why this is, though. RuboCop specs for the `rubocop` repo itself run quickly. I don't know why they are slow in my repo. But whatever; I guess maybe it's not worth trying to figure out.

PR #6342, opened 2025-03-06
[comments] Render Google login/signup button inline [COM-2]

davidrunger:
(... rather than redirecting to standalone login page.)

Saving the user a click and the mental processing of an additional login page will be a better user experience.

image

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -178,11 +178,13 @@ class ApplicationController < ActionController::Base
   # rubocop:enable Metrics/PerceivedComplexity
 
   def redirect_location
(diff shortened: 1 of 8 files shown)


app/controllers/application_controller.rb
```diff
@@ -178,11 +178,13 @@ def prioritize_json_format
   # rubocop:enable Metrics/PerceivedComplexity
 
   def redirect_location
-    next_redirect_chain_value || session.delete('user_return_to') || root_path
+    next_redirect_chain_value ||
+      session.delete('user_return_to') ||
+      root_path
```
davidrunger, line 183: (I was going to add a fourth option, but didn't end up doing so, but thought it's easier to read split out, anyway, so I'm leaving this change.)

PR #6357, opened 2025-03-07
[GitHub Action] Configure database to be fast for tests

davidrunger:
It's hard to be sure whether this is having much of an effect, but it seems fairly unlikely to hurt anything, and it will probably help at least somewhat (and maybe significantly), so let's try it.

Private convo: https://x.com/i/grok?conversation=1898060880509637048

---

Based on common PostgreSQL performance tuning practices and the context of test environments, it’s likely that GitLab adjusts certain PostgreSQL settings in test mode to make their tests run faster. In test environments, where data durability is less critical than speed, settings that reduce disk I/O are often modified to optimize performance. Here are the PostgreSQL settings that GitLab most likely changes:

- `fsync=off`: This setting disables the forced synchronization of data to disk, which speeds up write operations by reducing the number of times PostgreSQL waits for the operating system to confirm that data has been written. While this sacrifices data integrity in the event of a crash, that’s typically acceptable in a test environment where data is disposable.
- `synchronous_commit=off`: When set to `off`, this allows transactions to be considered committed without waiting for the Write-Ahead Log (WAL) to be written to disk. This reduces the overhead of disk operations, further accelerating test execution.
- `full_page_writes=off`: Disabling this setting minimizes the amount of data written to the WAL by preventing full page images from being logged during writes. This can improve performance by lowering the logging overhead, though it increases the risk of data corruption in a crash—again, a trade-off that’s tolerable in testing.

These adjustments are particularly effective in test mode, such as in GitLab’s CI/CD pipelines, where databases are often recreated for each test run, making data durability a non-issue. To implement these changes, GitLab would need to configure them at the server level, either by modifying the `postgresql.conf` file or by passing command-line options when starting the PostgreSQL server, as these parameters cannot be altered after the database is initialized.

In summary, GitLab likely sets `fsync=off`, `synchronous_commit=off`, and `full_page_writes=off` in their PostgreSQL configuration for test mode to reduce disk I/O and make their tests run faster.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -85,6 +85,16 @@ jobs:
       - name: Ensure clean git status
         run: '! [[ -n $(git status --porcelain) ]] || (git status && false)'
 
+      - name: Configure database to be fast for tests
+        run: |
+          function set() {
+            docker exec ${{ job.services.postgres.id }} sh -c "echo $1=\'$2\' >> $PGDATADIR/postgresql.conf"
+          }
+          set fsync off
+          set full_page_writes off
+          set synchronous_commit off
+          docker kill --signal=SIGHUP ${{ job.services.postgres.id }}
+
       - name: Run tests & linters
         run: bin/run-tests
         env:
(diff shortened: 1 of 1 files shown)


davidrunger:
Completes https://davidrunger.atlassian.net/browse/DEV-205 .

PR #6365, opened 2025-03-10
Replace all direct 'kyApi' usage with 'http' wrapper [DEV-202]

davidrunger:
This will make it easier to swap out ky for another package (or just fetch) if we ever want to, and it will make the codebase more consistent / standardize it on our “best practice” approach. It also lets us drop a decent number of `json()` calls, since we do that automatically in the `http` wrapper methods.


lib/test/tasks/run_file_size_checks.rb
```diff
@@ -8,7 +8,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
   CONSTRAINTS = {
     'charts*.js' => (295..305),
     'check_ins*.css' => (10..15),
-    'check_ins*.js' => (191..201),
+    'check_ins*.js' => (197..207),
```
davidrunger, line 11: I'm slightly surprised that we have to increase any of these file size allowances, but whatever. It's not by much, and maybe we had already crept up near the upper limit, anyway.

PR #6380, opened 2025-03-10
Only include spec-related rulesets when linting 'spec/'

davidrunger:
I'm hoping that this might improve performance a little bit? Even if it doesn't, it still feels like a good idea, so that we don't have cops that are only relevant to specs also linting non-spec code.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,12 +1,8 @@
 inherit_gem:
   runger_style:
-    - rulesets/capybara.yml
     - rulesets/default.yml
-    - rulesets/factory_bot.yml
     - rulesets/performance.yml
     - rulesets/rails.yml
-    - rulesets/rspec_rails.yml
-    - rulesets/rspec.yml
 
 require:
   - ./tools/custom_cops/require_all_custom_cops.rb
@@ -15,6 +11,7 @@ AllCops:
   Exclude:
     - !ruby/regexp /db\/migrate\/(201|202[0-4])/
   StringLiteralsFrozenByDefault: true
+  SuggestExtensions: false
 
 CustomCops/DontRollBackDatamigration:
   Include:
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -15,6 +11,7 @@ AllCops:
   Exclude:
     - !ruby/regexp /db\/migrate\/(201|202[0-4])/
   StringLiteralsFrozenByDefault: true
+  SuggestExtensions: false
```
davidrunger, line 14: Adding this to avoid the following warning:

```
The following RuboCop extension libraries are installed but not loaded in config:
  * rubocop-capybara
  * rubocop-factory_bot
  * rubocop-rspec
  * rubocop-rspec_rails

You can opt out of this message by adding the following to your config (see https://docs.rubocop.org/rubocop/extensions.html#extension-suggestions for more options):
  AllCops:
    SuggestExtensions: false
```

davidrunger:
Based on my local testing, this might save about 1 second, decreasing the total rubocop run time from ~30 seconds to ~29 seconds.

However, there is a downside, which is that it seems that rubocop config `Include` paths are interpreted relative to the config file, rather than being interpreted relative to the root directory. So, a cop like this:

```
Rails/ResponseParsedBody:
  Description: Prefer `response.parsed_body` to custom parsing logic for `response.body`.
  Enabled: pending
  Safe: false
  VersionAdded: '2.18'
  VersionChanged: '2.19'
  Include:
    - spec/controllers/**/*.rb
    - spec/requests/**/*.rb
    - test/controllers/**/*.rb
    - test/integration/**/*.rb
```

ends up not matching any files because there aren't any files matching e.g. `spec/spec/requests/**/*.rb`.

That's just one example; there are other similarly affected cops. This isn't a huge deal, though, as there aren't a ton of such cops. But I don't really feel that it's worth a ~1-second savings on RuboCop runs. Closing.

PR #6386, opened 2025-03-12
Move from bullet to prosopite for N+1 query protection [DEV-201]

davidrunger:
https://github.com/charkost/prosopite

Initial motivation: `prosopite` is more starred in my [`github-stars-explorer.ts`](https://github.com/davidrunger/dotfiles/blob/cc98273c0a18bcbb6b32d20abca15e8317cf460c/tools/github-stars-explorer.ts).

After trying it out (i.e. this PR), I can understand why it is more highly starred. The error messages seem better, and it seems to detect more N+1 queries. Also, IIRC, `bullet` errors immediately on any N+1 query, whereas I think `prosopite` will collect multiple errors across a request, so that they can all be fixed at once.

One downside of `prosopite` is that we also have to add `pg_query` to the Gemfile, which in turn requires `google-protobuf`, which seems to take a while to compile its native extensions. But usually that will be cached, so not a huge deal.


app/controllers/check_ins_controller.rb
```diff
@@ -34,27 +34,23 @@ def ensure_marriage
...
 
     bootstrap_data[:partner_ratings_of_user] =
       if @check_in.submitted_by_both_partners?
-        NeedSatisfactionRatingSerializer.new(
-          check_in_need_satisfaction_ratings.
-            where(user: current_user.marriage.decorate.other_partner).
-            includes(:emotional_need).
-            order('emotional_needs.name'),
-        )
+        partner_ratings_of_user
```
davidrunger, line 53: This is a now a little less efficient in the `else` case, in which case we have extracted `partner_ratings_of_user` unnecessarily, but it might still be a performance win, on net, since we save a query in this `if` case where we do use the `partner_ratings_of_user`. But, otherwise, `prosopite` will complain.

app/controllers/my_account_controller.rb
```diff
@@ -4,8 +4,14 @@ class MyAccountController < ApplicationController
   def destroy
     @user =
       User.includes(
-        logs: %i[log_entries log_shares],
+        :auth_tokens,
```
davidrunger, line 7: Not all of these additions were required to get specs passing, and, even with all of these additions, I think that this will still fail for my user in development mode. I am going to try to iterate on this more in future PRs. Here's one relevant ticket: https://davidrunger.atlassian.net/browse/DEV-208 .

app/policies/item_policy.rb
```diff
@@ -8,7 +8,7 @@ def resolve
   end
 
   def update?
-    scope.exists?(id: item.id)
+    (item.user == @user) || !!(@user.spouse && (item.user == @user.spouse))
```
davidrunger, line 11: I think that this is a pretty good catch by Prosopite, and this is one of the factors that makes me see it as better than `bullet`.

config/initializers/prosopite.rb
```diff
@@ -0,0 +1,9 @@
+if Rails.env.local? && !IS_DOCKER
```
davidrunger, line 1: The Prosopite docs suggest using `config/environments/development.rb` and `config/environments/test.rb` (as we were doing with Bullet), but I think that this more unified approach makes more sense.

lib/test/requirements_resolver.rb
```diff
@@ -269,6 +269,9 @@ def validate_task_config_groups!
       !file_changed?('app/javascript/types/index.ts') &&
         !diff_mentions?('intersection')
     end,
+    Test::Tasks::StartPercy => proc do
+      ENV['PERCY_TOKEN'].blank?
+    end,
```
davidrunger, line 274: (Mostly unrelated change that just made it a little easier/smoother for me to test this PR on my local machine.)

spec/controllers/logs/uploads_controller_spec.rb
```diff
@@ -47,7 +47,7 @@
 
       it 'creates log entries' do
         expect do
-          Sidekiq::Testing.inline! do
+          with_inline_sidekiq do
```
davidrunger, line 50: In this PR, in `with_inline_sidekiq`, we are wrapping the block with `Prosopite.pause`, which is why we switch to it here.

spec/controllers/my_account_controller_spec.rb
```diff
@@ -22,7 +22,7 @@
       context 'when requesting deletion of another user' do
         let(:logged_in_user) { User.excluding(user_to_destroy).first! }
 
-        it 'does not delete the specified user', :no_bullet do
```
davidrunger, line 25: I'm not sure why I had this `no_bullet` meta tag. I guess it passes with Prosopite active. 🤷‍♂️

spec/features/home_spec.rb
```diff
@@ -30,7 +30,9 @@
 
     page.scroll_to(:bottom)
 
-    wait_for { Event.count }.to eq(event_count_before + 1)
+    Prosopite.pause do
+      wait_for { Event.count }.to eq(event_count_before + 1)
+    end
```
davidrunger, line 35: I find it ever so slightly annoying that Prosopite catches "N + 1" queries in specs. I wouldn't even call this a real "N + 1" query; it's just that the same query happens to be made multiple times. But it's not a big negative, and I guess maybe that's just the price to be paid for Prosopite's otherwise useful N + 1 query detection.

PR #6387, opened 2025-03-12
Add MarriageMembership model (not foreign keys on marriage) [DEV-208]

davidrunger:
Motivation: this allows us to define User#marriage as a proper Rails association (which we can eager load), rather than defining it as an instance method on User (which we cannot eager load). This will allow us to get rid (in a future PR) of hacks such as [this][1].

[1]: https://github.com/davidrunger/david_runger/blob/f99333e0b247f83dd4f23a6739f47e122b890662/app/models/user.rb#L57-L69

I think that this data model is also a little bit more elegant, robust, and versatile. It feels right. As one example, it allows us to more confidently/accurately track when the second partner joined the marriage (which currently we could only guess to maybe/probably be Marriage#updated_at). It's also more flexible; for example, this new model can accommodate polygamous marriages, which our previous model could not.

To do this as safely as possible, I will do it in two PRs. The first will be this PR, which includes a datamigration to copy marriages to the new join table model. Then, in a subsequent PR, we will drop the `partner_1_id` and `partner_2_id` columns from the `marriages` table and remove any code related to those dropped foreign key columns.

### DIFF
--- a/app/controllers/my_account_controller.rb
+++ b/app/controllers/my_account_controller.rb
@@ -7,6 +7,10 @@ class MyAccountController < ApplicationController
         :auth_tokens,
         :need_satisfaction_ratings,
         logs: [:log_shares, { log_entries: :log_entry_datum }],
+        marriage: {
+          check_ins: %i[check_in_submissions need_satisfaction_ratings],
+          emotional_needs: :need_satisfaction_ratings,
+        },
         quiz_participations: %i[quiz_question_answer_selections],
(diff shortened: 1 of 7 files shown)


davidrunger:
I ran the datamigration ([private link](https://davidrunger.com/admin/datamigration_runs/2)). It was successful. Logs: [private link](https://grafana.davidrunger.com/goto/B7WjgShHg?orgId=1).

PR #6408, opened 2025-03-12
Refactor based on RubyCritic guidance

davidrunger:
This is by no means comprehensive. I'm sure that it leaves a lot on the table. I just looked through some of the RubyCritic feedback and made some refactors that looked good/helpful to me.


davidrunger:
Finishes https://davidrunger.atlassian.net/browse/DEV-204 .

PR #6414, opened 2025-03-13
Monkeypatch 'rspec-wait' to call 'Prosopite.pause'

davidrunger:
This makes it so that the repeated calls (if the expectation is not initially satisfied) in a `wait_for` block don't get flagged as an N + 1 query by Prosopite.

### DIFF
--- a/config/initializers/monkeypatches/rspec_wait_handler.rb
+++ b/config/initializers/monkeypatches/rspec_wait_handler.rb
@@ -11,19 +11,21 @@ if defined?(RSpec::Wait::Handler) && defined?(Prosopite)
       Prosopite.pause do
         matcher = RSpec.configuration.clone_wait_matcher ? initial_matcher.clone : initial_matcher
 
+        grandfather_method = method(:handle_matcher).super_method.super_method
+
         answer =
           if (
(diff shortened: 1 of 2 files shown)


config/initializers/monkeypatches/rspec_wait_handler.rb
```diff
@@ -0,0 +1,51 @@
...
+    # rubocop:disable Metrics
+    def handle_matcher(target, initial_matcher, message, &block)
+      start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+
+      answer = nil
+
+      Prosopite.pause do
+        matcher = RSpec.configuration.clone_wait_matcher ? initial_matcher.clone : initial_matcher
+
+        grandfather_method = method(:handle_matcher).super_method.super_method
```
davidrunger, line 14: This is fairly ugly. I think there's probably a better way to do this. But I'm just going to go with this for now.

What's happening here is basically that, in the original method, `super` was called here. Since we are adding another entry to the inheritance chain with `prepend`, that `super` method (in the original code) is the "grandfather" method of this method.

PR #6420, opened 2025-03-13
Update NPM packages

davidrunger:
`pnpm update --latest && pnpm dedupe`


davidrunger:
Tailwind continues to have a bug. :-/ How could there still be a bug through so many releases? Are others not facing this bug, as well? Am I doing something wrong on my end, and there is not really a bug? Anyway, closing for now.

PR #6426, opened 2025-03-15
Use Tailwind Vite plugin (not PostCSS plugin) [DEV-211]

davidrunger:
See here: https://tailwindcss.com/docs/installation/using-vite . The Vite plugin is recommended as the first/default option on that docs page.

Also, it seems to be explicitly recommended by the Tailwind team:

> it's still recommended to use the Vite plugin with Vite projects

-- https://github.com/tailwindlabs/tailwindcss/issues/ 17149#issuecomment-2720686585

We can't get rid of `postcss` completely, because it's still depended upon by other packages in the repo (such as `postcss-html`, which I think that we need to apply stylelint to Vue SFCs), but we can at least remove it from our `package.json` and we are closer to maybe someday removing it completely.


app/assets/stylesheets/styles.css
```diff
@@ -1,18 +1,18 @@
...
+@import './animations/appear_vertically';
+@import './components/spinner';
+@import './components/toastify';
+@import './elements/input';
+@import './flash';
+@import './utils/flex';
+@import './utils/font_family';
+@import './utils/font_weight';
+@import './utils/opacity';
+@import './ui_framework/button';
```
davidrunger, line 15: This change seems to have been needed (we got errors on these imports, otherwise), but this was the only change needed.

PR #6427, opened 2025-03-15
Add pghero (again) [DEV-210]

davidrunger:
https://github.com/ankane/pghero/blob/master/guides/Rails.md

First added here: https://github.com/davidrunger/david_runger/commit/ebcbb80d

Then removed here: https://github.com/davidrunger/david_runger/commit/09d09eb8

The reasons for removal were sound at the time, but I think they aren't really applicable anymore, and I think that there is value in bringing pghero back (particularly, to monitor for missing indexes).

Removal reasons (and reason I don't think it's applicable anymore):

1. prints Ruby 2.7 warnings (I highly doubt that this is the case anymore)
2. not compatible with `rails` master (I am not using Rails master anymore and I am guessing `pghero` is compatible with the latest released Rails version)
3. not worth it; DB performance isn't a big problem for the app (I think we might have some missing indexes, particularly in ActiveAdmin queries, and there might be value in knowing about that; also, costs aren't as high, now (see below), so "worth it" is an easier bar to meet)
4. tracking tables use/waste rows in database (now that we are on Hetzner rather than Heroku, DB usage is not nearly as much of a concern as it was before)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,7 @@ RUN --mount=type=cache,sharing=private,target=/var/lib/apt/lists \
   build-essential git libpq-dev libyaml-dev unzip
 
 # Download skedjewel binary.
-ARG SKEDJEWEL_VERSION=v0.0.15
+ARG SKEDJEWEL_VERSION=v0.0.16
 RUN curl --fail -L "https://github.com/davidrunger/skedjewel/releases/download/$SKEDJEWEL_VERSION/skedjewel-$SKEDJEWEL_VERSION-linux" > skedjewel && \
   mkdir -p /app/bin && \
   mv skedjewel /app/bin/ && \
--- a/config/database.yml
+++ b/config/database.yml
@@ -89,4 +89,4 @@ test:
 production:
   <<: *default
   database: david_runger_production
-  url: <%= ENV.fetch('DATABASE_URL', nil) %>
+  url: <%= ENV.fetch('DATABASE_URL', Rails.env.production? ? fail('Missing DATABASE_URL!') : nil) %>
--- a/config/pghero.yml
+++ b/config/pghero.yml
@@ -1,0 +1,52 @@
+databases:
+  primary:
+    # Database URL
+    url: <%= ENV.fetch('PGHERO_DATABASE_URL', Rails.env.test? ? nil : fail('Missing PGHERO_DATABASE_URL')) %>
+
+    # System stats
+    # aws_db_instance_identifier: my-instance
+    # gcp_database_id: my-project:my-instance
+    # azure_resource_id: my-resource-id
+
+  # Add more databases
+  # other:
+  #   url: <%= ENV["OTHER_DATABASE_URL"] %>
+
+# Minimum time for long running queries
+# long_running_query_sec: 60
+
+# Minimum average time for slow queries
+# slow_query_ms: 20
+
+# Minimum calls for slow queries
+# slow_query_calls: 100
+
+# Minimum connections for high connections warning
+# total_connections_threshold: 500
+
+# Explain functionality
+# explain: true / false / analyze
+
+# Statement timeout for explain
+# explain_timeout_sec: 10
+
+# Visualize URL for explain
+# visualize_url: https://...
+
+# Time zone (defaults to app time zone)
+# time_zone: "Pacific Time (US & Canada)"
+
+# Basic authentication
+# username: admin
+# password: <%= ENV["PGHERO_PASSWORD"] %>
+
+# Stats database URL (defaults to app database)
+# stats_database_url: <%= ENV["PGHERO_STATS_DATABASE_URL"] %>
+
+# AWS configuration (defaults to app AWS config)
+# aws_access_key_id: <%= ENV["AWS_ACCESS_KEY_ID"] %>
+# aws_secret_access_key: <%= ENV["AWS_SECRET_ACCESS_KEY"] %>
+# aws_region: us-east-1
+
+# Filter data from queries (experimental)
+# filter_data: true
--- a/config/skedjewel.yml
+++ b/config/skedjewel.yml
@@ -6,6 +6,9 @@ config:
 jobs:
   DataMonitors::Launcher: '**:07' # hourly at 7 minutes after
   SendLogReminderEmails: '**:**' # every minute
+  PgHero::CaptureQueryStats: '**:%5' # every 5 minutes
+  PgHero::CaptureSpaceStats: '22:52' # daily at 10:52pm Central Time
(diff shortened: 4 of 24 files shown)


Dockerfile
```diff
@@ -40,7 +40,7 @@ RUN --mount=type=cache,sharing=private,target=/var/lib/apt/lists \
   build-essential git libpq-dev libyaml-dev unzip
 
 # Download skedjewel binary.
-ARG SKEDJEWEL_VERSION=v0.0.15
+ARG SKEDJEWEL_VERSION=v0.0.16
```
davidrunger, line 43: https://github.com/davidrunger/skedjewel/releases/tag/v0.0.16

We need this version bump to schedule a PgHero Sidekiq job every 5 minutes (`PgHero::CaptureQueryStats: '**:%5'`).

config/database.yml
```diff
@@ -89,4 +89,4 @@ test:
 production:
   <<: *default
   database: david_runger_production
-  url: <%= ENV.fetch('DATABASE_URL', nil) %>
+  url: <%= ENV.fetch('DATABASE_URL', Rails.env.production? ? fail('Missing DATABASE_URL!') : nil) %>
```
davidrunger, line 92: (Mostly unrelated change, but an opportunity for improvement that I saw while looking at this.)

config/skedjewel.yml
```diff
@@ -6,6 +6,9 @@ config:
 jobs:
   DataMonitors::Launcher: '**:07' # hourly at 7 minutes after
   SendLogReminderEmails: '**:**' # every minute
+  PgHero::CaptureQueryStats: '**:%5' # every 5 minutes
```
davidrunger, line 9: (See https://github.com/davidrunger/david_runger/pull/6427/files#r1997352708 .)

db/migrate/20171223044916_create_pghero_query_stats_old.rb
```diff
@@ -1,4 +1,4 @@
-class CreatePgheroQueryStats < ActiveRecord::Migration[5.1]
+class CreatePgheroQueryStatsOld < ActiveRecord::Migration[5.1]
```
davidrunger, line 1: I feel sort of bad changing these file names and class names, but otherwise Rails was complaining about the migrations already existing.

db/migrate/20171223045057_create_pghero_space_stats_old.rb
```diff
@@ -1,4 +1,4 @@
-class CreatePgheroSpaceStats < ActiveRecord::Migration[5.1]
+class CreatePgheroSpaceStatsOld < ActiveRecord::Migration[5.1]
```
davidrunger, line 1: (See https://github.com/davidrunger/david_runger/pull/6427/files#r1997353937 .)

db/migrate/20250315174325_create_pghero_query_stats.rb
```diff
@@ -0,0 +1,17 @@
+class CreatePgheroQueryStats < ActiveRecord::Migration[8.0]
+  def change
+    # rubocop:disable Rails/CreateTableWithTimestamps
```
davidrunger, line 3: There's a `captured_at` timestamp, which I guess is enough. I don't want to depart from PgHero's expected data model.

db/migrate/20250315193051_create_pghero_space_stats.rb
```diff
@@ -0,0 +1,15 @@
+class CreatePgheroSpaceStats < ActiveRecord::Migration[8.0]
+  def change
+    # rubocop:disable Rails/CreateTableWithTimestamps
```
davidrunger, line 3: There's a `captured_at` timestamp, which I guess is enough. I don't want to depart from PgHero's expected data model.

db/migrate/20250315194328_enable_pg_stat_statements_extension_v2.rb
```diff
@@ -0,0 +1,9 @@
+class EnablePgStatStatementsExtensionV2 < ActiveRecord::Migration[8.0]
+  def up
+    ApplicationRecord.connection.execute('CREATE EXTENSION pg_stat_statements;')
+  end
+
+  def down
+    ApplicationRecord.connection.execute('DROP EXTENSION pg_stat_statements;')
+  end
```
davidrunger, line 8: I'm not sure where I came up with the idea for this migration when I added `pghero` last time. It seems that the docs don't suggest it, and instead maybe there's a button in the PgHero web UI to click to enable this extension? But adding it in a migration and having it in the schema seems like a good way to go.

docker-compose.yml
```diff
@@ -186,6 +186,16 @@ services:
...
+    command:
+      - postgres
+      - -c
+      - shared_preload_libraries=pg_stat_statements
+      - -c
+      - pg_stat_statements.track=all
+      - -c
+      - pg_stat_statements.max = 10000
+      - -c
+      - track_activity_query_size = 2048
```
davidrunger, line 198: This is a modification on https://github.com/ankane/pghero/blob/master/guides/Query-Stats.md#installation 

> Add the following to your `postgresql.conf`:
>
> ```
> shared_preload_libraries = 'pg_stat_statements'
> pg_stat_statements.track = all
> pg_stat_statements.max = 10000
> track_activity_query_size = 2048
> ```

script/create_pghero_user.rb
```diff
@@ -0,0 +1,74 @@
...
+require 'uri'
+
+url = ENV.fetch('PGHERO_DATABASE_URL')
+uri = URI.parse(url)
+pghero_password = uri.password
+database_name = uri.path.delete_prefix('/')
+main_db_user = 'david_runger'
+
+ApplicationRecord.connection.execute(<<~SQL)
+  CREATE SCHEMA pghero;
```
davidrunger, line 16: Tweaked very slightly (mostly just filling in the blanks) from here https://github.com/ankane/pghero/blob/master/guides/Permissions.md .

config/pghero.yml
```diff
@@ -0,0 +1,52 @@
+databases:
+  primary:
+    # Database URL
+    url: <%= ENV.fetch('PGHERO_DATABASE_URL', Rails.env.test? ? nil : fail('Missing PGHERO_DATABASE_URL')) %>
```
davidrunger, line 4: Requiring this in development or production also causes problems. Fixed (hopefully) here: https://github.com/davidrunger/david_runger/pull/6428 .

config/database.yml
davidrunger: Actually, it wasn't a good idea. Fixed here: https://github.com/davidrunger/david_runger/pull/6429 .

~~I feel like I might have tried to do this before...~~ **Edit:** Looking at `fcom DATABASE_URL -p config/database.yml`, I guess I haven't tried to do this before.

PR #6428, opened 2025-03-15
Don't require PGHERO_DATABASE_URL

davidrunger:
I believe that requiring it was the cause of this deploy failure: https://github.com/davidrunger/david_runger/actions/runs/13877192439/job/38830995512#step:5:167

Also, add `create_schema "pghero"` to `db/schema.rb`. This happened after `dbm` was run on my machine after having run `ruby script/create_pghero_user.rb`. I guess it's a good thing to commit.

### DIFF
--- a/config/pghero.yml
+++ b/config/pghero.yml
@@ -1,7 +1,7 @@
 databases:
   primary:
     # Database URL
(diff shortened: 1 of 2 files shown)


davidrunger:
If we try to use PgHero without this set, then we get an error, which is good:

image

But now I think we won't get an error when, for example, just running migrations or precompiling assets.

PR #6430, opened 2025-03-15
Remove spaces from postgres configs

davidrunger:
The site is currently down because Postgres is failing to boot because of this.

```
docker compose logs postgres --since 3m
postgres-1  | 
postgres-1  | PostgreSQL Database directory appears to contain a database; Skipping initialization
postgres-1  | 
postgres-1  | 2025-03-15 22:51:13.723 GMT [1] FATAL:  invalid configuration parameter name "pg_stat_statements.max "
postgres-1  | 2025-03-15 22:51:13.723 GMT [1] DETAIL:  Custom parameter names must be two or more simple identifiers separated by dots.
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -193,9 +193,9 @@ services:
       - -c
       - pg_stat_statements.track=all
       - -c
-      - pg_stat_statements.max = 10000
+      - pg_stat_statements.max=10000
       - -c
-      - track_activity_query_size = 2048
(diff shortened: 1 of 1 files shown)


davidrunger:
I guess that I hadn't checked that `LOCAL_TEST=1 bin/server/deploy.sh` succeeds with the latest version of this, because, testing it just now (against `main` without this PR in it), it does indeed fail. Bad form.

PR #6432, opened 2025-03-16
Pin all GitHub Actions to a particular SHA [DEV-212]

davidrunger:
This will make us less vulnerable to supply chain attacks like this one: https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised .

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -52,7 +52,8 @@ jobs:
           PGUSER: postgres
           PGPASSWORD: postgres
 
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Check out code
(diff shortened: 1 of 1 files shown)


.github/workflows/ruby.yml
```diff
@@ -52,7 +52,8 @@ jobs:
           PGUSER: postgres
           PGPASSWORD: postgres
 
-      - uses: actions/checkout@v4
+      - name: Check out code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
davidrunger, line 56: I guess that version comments in this format will allow Dependabot to update them? Per https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash :

> Adding a comment referring to the tag enables automated dependency updates, such as Dependabot or Renovatebot, to update it whenever a new release is detected. [You can see an example of this in action in the Node.js Security WG repository](https://github.com/nodejs/security-wg/pull/1009).

PR #6433, opened 2025-03-16
Leverage eager loading when destroying user -> marriage [DEV-213]

davidrunger:
This avoids some N + 1 queries when a user deletes their account.

I believe that all of the remaining N + 1 queries are those (seemingly unnecessarily) caused by `paper_trail`, as mentioned [here][1].

[1]: https://davidrunger.atlassian.net/browse/DEV-213?focusedCommentId=10704

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -61,8 +61,8 @@ class User < ApplicationRecord
   has_paper_trail_for_all_events
 
   before_destroy do |user|
-    if (marriage_id = user.marriage_membership&.marriage_id)
-      Marriage.find(marriage_id).destroy!
+    if (marriage = user.marriage)
+      marriage.destroy!
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -61,8 +61,8 @@ class User < ApplicationRecord
   has_paper_trail_for_all_events
 
   before_destroy do |user|
-    if (marriage_id = user.marriage_membership&.marriage_id)
-      Marriage.find(marriage_id).destroy!
+    if (marriage = user.marriage)
+      marriage.destroy!
```
davidrunger, line 65: We eager load this marriage and some of its associations here: https://github.com/davidrunger/david_runger/blob/b8774fc3659c3029e2ff5f10062527c03e65d0f7/app/controllers/my_account_controller.rb#L10-L13

This change leverages that eager loading.

I'm not sure why we were previously doing it the way that we were. 😕

PR #6435, opened 2025-03-17
Render flash messages for home page via vue-toastification [DEV-214]

davidrunger:
**Motivation:** HTML-rendered flash messages are at least partially/largely overwritten when a Vue app renders into the page, so it looks pretty bad. This way, it won't look so bad.

This change also updates toast messages for the logs app to be rendered by `vue-toastification` rather than by `toastify`. This is for standardization and because the `vue-toastification` toasts have a butter appearance and a better UX.

The file size increases (JS and CSS) for the home app and the logs app are not terrible.

Before | After
|-|-|
image | image


app/controllers/home_controller.rb
```diff
@@ -2,6 +2,7 @@ class HomeController < ApplicationController
   include Prerenderable
 
   skip_before_action :authenticate_user!, only: %i[index upgrade_browser]
+  render_flash_messages_via_js(only: %i[index])
```
davidrunger, line 5: We might want to do this for some other controller actions, in the future, but for now the real concern is for home#index because that is the fallback/default location for redirects after an error of some kind.

app/javascript/lib/toasts.ts
```diff
@@ -22,14 +20,3 @@ export function toast(message: string, options?: ToastifyOptions) {
     className: options?.type && TYPE_TO_CLASS_MAP[options?.type],
   }).showToast();
 }
-
-export function renderBootstrappedToasts() {
```
davidrunger, line 26: This is moved to `vue_toasts.ts` (and now using `vue-toastification` rather than `toastify`).

PR #6480, opened 2025-03-22
Update NPM packages

davidrunger:
`pnpx npm-check-updates --upgrade && pnpm install && pnpm dedupe`


davidrunger:
Closing because the only update is for Tailwind which still has a bug.

PR #6491, opened 2025-03-23
[api/logs#destroy] Fix N+1 query

davidrunger:
Prosopite flagged the N+1 query when I deleted a log locally to test a change for https://davidrunger.atlassian.net/browse/DEV-217 .

I wonder if `bullet` would also have caught this? It would seem that it didn't, but we didn't have test coverage for it before, and it's possible that maybe I just never deleted a log in development since adding `bullet`. That seems slightly unlikely, though. I'm guessing that `bullet` had a chance to catch this, but failed. If so, then, nice job, Prosopite, catching what Bullet missed.

### DIFF
--- a/spec/controllers/api/logs_controller_spec.rb
+++ b/spec/controllers/api/logs_controller_spec.rb
@@ -122,13 +122,14 @@ RSpec.describe Api::LogsController do
     context "when attempting to destroy one's own log" do
       let(:user) { log.user }
 
-      it 'destroys the log' do
-        expect { delete_destroy }.to change { Log.find_by(id: log.id) }.from(Log).to(nil)
-      end
+      context 'when the log has more than one entry' do
+        before { expect(log.log_entries.size).to be > 1 }
 
-      it 'returns a 204 status code' do
-        delete_destroy
-        expect(response).to have_http_status(204)
+        it 'destroys the log and returns a 204 status code without raising an N+1 query error' do
+          expect { delete_destroy }.to change { Log.find_by(id: log.id) }.from(Log).to(nil)
+
+          expect(response).to have_http_status(204)
+        end
       end
     end
   end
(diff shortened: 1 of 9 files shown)


spec/models/log_spec.rb
```diff
@@ -24,7 +24,7 @@
 
         context 'when all log entries were created long enough ago that a reminder is needed' do
           before do
-            log.log_entries.find_each do |log_entry|
+            log.log_entries.includes(:log_entry_datum).find_each do |log_entry|
```
davidrunger, line 27: It's a tiny bit annoying that Prosopite enforces eager loading in spec code, not just application code, but at the same time it's maybe/probably somewhat of a good thing. For one thing, spec performance does matter, too (although, in most cases, the number of records involved will probably make eager loading a relatively negligible cost).

spec/controllers/api/logs_controller_spec.rb
```diff
@@ -122,13 +122,14 @@
...
+        before { expect(log.log_entries.size).to be > 1 }
 
-      it 'returns a 204 status code' do
-        delete_destroy
-        expect(response).to have_http_status(204)
+        it 'destroys the log and returns a 204 status code without raising an N+1 query error' do
+          expect { delete_destroy }.to change { Log.find_by(id: log.id) }.from(Log).to(nil)
+
+          expect(response).to have_http_status(204)
+        end
```
davidrunger, line 132: Here I am combining two examples into one. I used to be somewhat religious, at least in unit specs, about having only one expectation per example. However, in feature specs, for performance reasons, we combine lots of expectations into a single example, and really it doesn't seem to cause much (or really almost any) difficulty. So, I think I'm going to probably start combining some more unit tests together, too, when it feels appropriate. This seems to me like a decent occasion for consolidation.

PR #6505, opened 2025-03-24
[home/projects] Remove deprecated 'slot="overview"' attributes

davidrunger:
I think this is not doing anything, anyway. The `template(#overview)` outside each of the divs being removed herein is what causes the content to be rendered into the `overview` slot.


davidrunger:
# View ignoring whitespace changes

https://github.com/davidrunger/david_runger/pull/6505/files?w=1

PR #6509, opened 2025-03-24
[EditLogSharingSettingsModal] Drop 'native' event modifier

davidrunger:
Also, while we're here, fix some styling issues.

| Before | After |
| - | - |
| image | image |


davidrunger:
The `native` modifier is deprecated. That's the motivation for this change.

PR #6522, opened 2025-03-25
Render Google SSO button/form as a web component (not PNG) [DEV-220]

davidrunger:
I saw a somewhat blurry sign-in button on my iPhone when not signed in and looking at the comments section of a blog article.

I think that rendering the button as a web component (with actual HTML "Sign in with Google" text and an SVG Google icon, rather than a PNG image of that text and Google icon) is a nice way to deliver the button along with its associated styles, for use in both Rails views and Vue components.

Unfortunately, the testing situation with custom elements / shadow DOM seems to be pretty bleak. I added the `capybara-shadowdom` gem, but it doesn't work really smoothly, and I sort of had to hack a solution in the `click_sign_in_with_google` spec helper method that I'm adding in this change.


app/javascript/custom_elements/google_sign_in_button.ts
```diff
@@ -0,0 +1,155 @@
...
+            box-shadow: 0 1px 2px 0 rgba(60, 64, 67, .30), 0 1px 3px 1px rgba(60, 64, 67, .15);
+          }
+
+          .gsi-material-button:not(:disabled):hover .gsi-material-button-state {
+            background-color: #303030;
+            opacity: 8%;
+          }
+        </style>
+
+        <form class="gsi-form" method="post">
```
davidrunger, line 121: Note that (despite the name) it's not only a button, but also a form.

davidrunger:
It looks notably better. (View at full resolution).

| Before | After |
| - | - |
| IMG_2535 | IMG_2534 |

One unanticipated side benefit is that, on my mobile device, the "Sign in with Google to add a comment." text now stays on a single line, rather than wrapping.

PR #6526, opened 2025-03-25
Don't save admin login page URL as redirect_chain

davidrunger:
I noticed this bug:

1. clear cookies
2. log in to admin
3. try to go to a user-facing page
4. be redirected to log in; do so

**Expect:** redirected to the user page that I was trying to access in step 3.

**Actual:** redirected to the admin dashboard.

This change fixes that bug. Now the User will be redirected to the User-facing page that they had tried to access, as expected and intended.

### DIFF
--- a/spec/features/user_google_login_spec.rb
+++ b/spec/features/user_google_login_spec.rb
@@ -48,6 +48,39 @@ RSpec.describe 'Logging in as a User via Google auth', :prerendering_disabled do
               end
             end
           end
+
+          context 'when no User or AdminUser is logged in' do
+            before { Devise.sign_out_all_scopes }
+
+            context 'when there is an AdminUser and a User with the same stubbed_user_email' do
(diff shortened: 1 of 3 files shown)


spec/features/user_google_login_spec.rb
```diff
@@ -48,6 +48,39 @@
...
+                context 'when the user tries to go to the My Account page' do
+                  it 'redirects the user to log in and, after successful login, redirects to the My Account page' do
+                    visit(my_account_path)
+
+                    expect(page).to have_current_path(new_user_session_path)
+
+                    click_on('Sign in with Google')
+
+                    expect(page).to have_current_path(my_account_path)
+                    expect(page).to have_text('My Account')
```
davidrunger, line 78: Prior to this change, this spec would fail by redirecting the User to `/admin` after logging in as a User to access a User-requiring route. Now, we redirect to the intended User page (My Account, in this example), as intended.

PR #6537, opened 2025-03-25
[ApplicationPolicy] Require a user for create? and new?

davidrunger:
We usually want to know who has created the relevant record and who owns it.

There are some cases where we now need to pass an AuthToken-authenticated user, which is accomplished via the new `allow_auth_token_authorization` controller class method.

There are some exceptions to this, which are reflected by the changes herein, where, for some specific policies (EventPolicy, UserPolicy, WebhookEmailForwardPolicy), we continue with `created?` always returning `true` (even if there is no user).

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -17,7 +17,7 @@ class ApplicationPolicy
   end
 
   def create?
-    true
+    @user.is_a?(User)
   end
 
   def new?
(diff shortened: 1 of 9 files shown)


app/policies/application_policy.rb
```diff
@@ -17,7 +17,7 @@ def show?
   end
 
   def create?
-    true
+    @user.is_a?(User)
```
davidrunger, line 20: Here is the key change. Pretty much all other changes in this PR are downstream from this.

PR #6573, opened 2025-03-28
[comments] Render dark Google SSO button if user prefers [COM-8]

davidrunger:
It doesn't look amazing, and just on pure aesthetics I think it looks a bit worse than a light mode SSO button. For this reason, I was going to not go ahead with this. However, a downside of the light mode SSO button is that it stands out sort of a lot and it is kind of a visual distraction. It can draw one's eye away from the article text as one gets near the end of the article. That's why I am going ahead with this change.

| Before | After |
| - | - |
| image | image |

### DIFF
--- a/app/javascript/comments/Comments.vue
+++ b/app/javascript/comments/Comments.vue
@@ -5,7 +5,7 @@
   template(v-if="store.currentUser")
     CommentForm(@submit="store.addComment")
   template(v-else)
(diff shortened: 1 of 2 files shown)


davidrunger:
> It doesn't look amazing, and just on pure aesthetics I think it looks a bit worse than a light mode SSO button

This is basically rectified in #6574.

PR #6575, opened 2025-03-28
Migrate from Vuelidate to Regle [DEV-226]

davidrunger:
Per this issue https://github.com/vuelidate/vuelidate/issues/ 1163 and the fact that Vuelidate hasn't had any releases since June 2023, the project is sort of dead. https://github.com/victorgarciaesgi/regle aims to be its successor, keeping a similar API and adding better TypeScript support, and also presumably being actively/better maintained. So, this change moves us from Vuelidate to Regle.


davidrunger:
For now, I can't bring myself to do this. Vuelidate still works. It's also smaller. I'm also not really leveraging it very much and I don't plan to leverage it a lot more in the future, so it seems less critical to be on a well maintained library, since it won't be so hard to change (e.g. resurrect this PR) if something in Vuelidate ever breaks.

According to https://github.com/vuelidate/vuelidate/network/dependents , Vuelidate is used by ~48k packages. According to https://github.com/victorgarciaesgi/regle/network/dependents , Regle is used by ~26 repositories. This is many fewer eyes on the code, and many fewer tools scanning for vulnerabilities.

I would like to help Regle to grow, but for now it seems a bit riskier, and it's hard for me to justify that risk (and the ~29 KiB of extra page weight). I'm going to close this, for now.

It's good to know that Regle is an option, though, in case I ever run into issues with Vuelidate in the future, or in case Vuelidate loses compatibility with the latest versions of Vue.

PR #6588, opened 2025-03-29
Use modal store to track keydown listener registration [DEV-232]

davidrunger:
... rather than using globally accessible and mutatable `window.davidrunger`.

Also, add mostly unrelated `ElTooltip` import (noticed while testing the primary change).

Also, add mostly unrelated `type` for `Log` import (which I noticed the absence of was breaking the logs app in development when testing).

### DIFF
--- a/app/javascript/groceries/components/CheckInItem.vue
+++ b/app/javascript/groceries/components/CheckInItem.vue
@@ -34,7 +34,7 @@ li.flex.items-center.break-word.mb-2(:class="aboutToMoveToClass()")
 </template>
 
 <script setup lang="ts">
-import { ElButton } from 'element-plus';
+import { ElButton, ElTooltip } from 'element-plus';
 import { storeToRefs } from 'pinia';
 import { HeartFilledIcon } from 'vue-tabler-icons';
 import { object } from 'vue-types';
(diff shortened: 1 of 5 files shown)


app/javascript/groceries/components/CheckInItem.vue
```diff
@@ -34,7 +34,7 @@ li.flex.items-center.break-word.mb-2(:class="aboutToMoveToClass()")
 </template>
 
 <script setup lang="ts">
-import { ElButton } from 'element-plus';
+import { ElButton, ElTooltip } from 'element-plus';
```
davidrunger, line 37: > Also, add mostly unrelated `ElTooltip` import (noticed while testing the primary change).

app/javascript/logs/components/NewLogEntryForm.vue
```diff
@@ -48,7 +48,7 @@ import { object } from 'vue-types';
 
 import { isArrayOfNumbers } from '@/lib/type_predicates';
 import { useLogsStore } from '@/logs/store';
-import { Log } from '@/logs/types';
+import { type Log } from '@/logs/types';
```
davidrunger, line 51: > Also, add mostly unrelated `type` for `Log` import (which I noticed the absence of was breaking the logs app in development when testing).

PR #6599, opened 2025-03-30
[comments] Load styles.css so that we can use Tailwind [COM-9]

davidrunger:
It feels like sort of a weird quirk (and it's also sort of annoying / a lost opportunity) that, prior to this change, I could not use Tailwind classes to style components in the comments app. This change fixes that by injecting the styles.css style sheet into blog #show pages (the only ones where we currently render the comments app). This will add some unnecessary page weight, but I guess that's true of every page where we load the Tailwind styles, so it's not a new/unique problem, and styles.css is currently compiling to just 20.2 kB uncompressed and 6.3 kB gzipped, which is not so much, and also I believe that CSS is not as costly to the browser to process as JavaScript, compared byte-for-byte.

### DIFF
--- a/app/assets/stylesheets/tailwind_overrides.css
+++ b/app/assets/stylesheets/tailwind_overrides.css
@@ -86,10 +86,5 @@
     select {
       @apply border;
     }
-
-    input,
-    textarea {
-      background-color: revert;
-    }
   }
 }
--- a/app/controllers/blog_controller.rb
+++ b/app/controllers/blog_controller.rb
@@ -67,21 +67,7 @@ class BlogController < ApplicationController
     )
       send_file(absolute_path, **kwargs)
     elsif starts_with_blog_directory && absolute_path.end_with?('.html')
-      # rubocop:disable Rails/OutputSafety
-      render(
-        html:
-          File.read(absolute_path).
-            sub(
-              '</head>',
-              "#{helpers.csrf_meta_tags}#{helpers.window_data_script_tag}</head>",
-            ).
-            sub(
-              '</body>',
-              "#{helpers.ts_tag('comments')}</body>",
-            ).
-            html_safe,
-      )
-      # rubocop:enable Rails/OutputSafety
+      render(html: blog_html_with_additions(absolute_path))
     else
       Rails.error.report(
         Error.new(UnauthorizedBlogFileRequest),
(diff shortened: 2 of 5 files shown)


app/assets/stylesheets/tailwind_overrides.css
```diff
@@ -86,10 +86,5 @@
     select {
       @apply border;
     }
-
-    input,
-    textarea {
-      background-color: revert;
-    }
```
davidrunger, line 93: This was added in https://github.com/davidrunger/david_runger/pull/6495 . However, I think that Tailwind's default -- a transparent background -- actually is pretty good, in most cases. On a dark background, we want the textarea to also be dark, not white (which is what `revert` achieves).

In other words, on closer reflection, it turns out that this is just something that we want for the groceries app, because it doesn't have dark mode and because it has an image background.

So, this change moves it to the (unscoped) style tag in `Groceries.vue`.

app/controllers/blog_controller.rb
```diff
@@ -107,6 +93,31 @@ def send_blog_file(relative_path, **kwargs)
...
+      '</body>',
+      "#{helpers.ts_tag('comments')}</body>",
+    )
+
+    if params[:action] == 'show'
+      html.sub!(
+        '</head>',
+        "#{helpers.ts_tag('styles')}</head>",
+      )
+    end
```
davidrunger, line 114: Here's the new part; the rest of this method is essentially just extracted from a different location (with a slight tweak to use `sub!` rather than `sub`).

app/javascript/comments/Comments.vue
```diff
@@ -7,7 +7,7 @@
   template(v-else)
     p #[GoogleLoginButton.google-login-form.dark-mode-supported(:origin="googleLoginOrigin")] to add a comment.
 
-  .comments
+  .mt-8
```
davidrunger, line 10: This was sort of just a proof of concept / a test. I think that there are probably other places where we could use Tailwind classes in the comments app, now that we can. But, to keep scope narrow, I didn't look for any other places to do that, apart from this spot here.

app/javascript/comments/Comments.vue
```diff
@@ -50,19 +50,11 @@ const googleLoginOrigin = [
 </script>
 
 <style scoped>
-.google-login-form {
-  display: inline;
-}
```
davidrunger, line 55: I was initially going to do the test replacing this class, but it seems not to be needed (anymore?).

app/javascript/groceries/Groceries.vue
```diff
@@ -129,4 +129,9 @@ main {
     font-size: 16px;
   }
 }
+
+input,
+textarea {
+  background-color: revert;
+}
```
davidrunger, line 136: Ref: https://github.com/davidrunger/david_runger/pull/6599/files#r2020033696

spec/controllers/blog_controller_spec.rb
```diff
@@ -41,19 +41,20 @@
...
             end
           end
 
-          it 'responds with 200 and the show file content' do
+          it 'responds with 200 and the show file content (with additions, such as a styles.css stylesheet tag)' do
             get_show
 
             expect(response).to have_http_status(200)
-            expect(response.body).to eq(show_content)
+            expect(response.body).to match(/\bstyles\b/)
```
davidrunger, line 56: This test is slightly awkward / not tightly specified because, in prod, I think that this will be a stylesheet, but in specs I think it might be a script tag for `styles.ts`. So, I'm just checking for `styles`.

spec/controllers/blog_controller_spec.rb
```diff
@@ -41,19 +41,20 @@
         let(:format) { nil }
 
         context 'when the requested file exists in the blog/ directory' do
-          let(:show_content) { 'it is the best language ever' }
+          let(:show_content) { '</head>it is the best language ever' }
```
davidrunger, line 44: Add `</head>` so that there is something for the `blog_html_with_additions` method to "hook into".

PR #6603, opened 2025-03-30
Update NPM packages

davidrunger:
`pnpx npm-check-updates --upgrade && pnpm install && pnpm update && pnpm dedupe`


davidrunger:
This tries to upgrade Tailwind, which we don't want / which breaks things.

PR #6605, opened 2025-03-30
Bump vega, vega-embed, and add explicitly and bump vega-lite

davidrunger:
1. `vega` 5.33.0 to 6.1.2
2. `vega-embed` 6.29.0 to 7.0.2
3. `vega-lite` _(indirect)_ 5.23.0 to _(explicit)_ 6.1.0

Everything seems to still work.

I'm adding `vega-lite` explicitly because otherwise it seems hard (or, at least, not as easy) to bump the version and/or to notice if the version is not the latest (via `pnpm outdated`). This is similar to #6583, which added `vega` explicitly.

I'm changing the `tsconfig` `moduleResolution` (docs: https://www.typescriptlang.org/tsconfig/#moduleResolution ) from `node` to `bundler` because it avoids this error:

```
Running system command './node_modules/.bin/vue-tsc --noEmit' with ENV vars {} ...
app/javascript/ci_step_results/renderGanttCharts.ts(1,47): error TS2307: Cannot find module 'vega-embed' or its corresponding type declarations.
  There are types at '/home/david/code/david_runger/node_modules/vega-embed/build/embed.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
'./node_modules/.bin/vue-tsc --noEmit' with ENV vars {} failed (exited with 1, took 18.866).
```

The `node` option we were using before apparently refers to `node10`, which I think is quite old, so I think that this is overdue for an update. I am not sure if `bundler` is what we want, but it probably is, since we are using Vite.


lib/test/tasks/run_file_size_checks.rb
```diff
@@ -10,7 +10,7 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'check_ins*.css' => (10..15),
     'check_ins*.js' => (197..207),
     'check_ins_index*.js' => (5..15),
-    'ci_step_gantt_charts*.js' => (830..840),
+    'ci_step_gantt_charts*.js' => (806..816),
```
davidrunger, line 13: It's always nice to see a decrease (though, in absolute terms, this file is still pretty huge).

I think that this might possibly be because there is no longer duplication in one of the vega packages, but I'm not very sure about that at all.

PR #6608, opened 2025-03-30
Lower RAILS_MAX_THREADS from 2 to 1 except for web

davidrunger:
**Motivation:** Even with our 2 GiB of memory on Hetzner, swap usage continues to be a bit of a problem. I'm hoping that this change (decreasing the RAILS_MAX_THREADS from 2 to 1 for the `worker` service) will decrease memory usage by the `worker` dyno to a helpful extent. Really low Sidekiq latency is not important for the app, so lowering the Sidekiq thread count to 1 should not be a problem.

As part of this change, move `PROMETHEUS_EXPORTER_HOST` to the `Dockerfile`, so that we don't have to worry about overwriting it (problematic) or redefining it (annoying) in the `web` service when we override the `environment` section.

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -1,3 +1,2 @@
-RAILS_MAX_THREADS=2
 # https://island94.org/2025/01/ruby-thread-contention-simply-gvl-queuing
 RUBY_THREAD_TIMESLICE=10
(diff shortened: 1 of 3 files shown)


davidrunger:
This appears to have maybe had a small effect in decreasing worker memory usage, but not a ton?

https://grafana.davidrunger.com/goto/0DjxRjTNg?orgId=1

image

The cursor is hovering over the approximate time of the deploy of this PR.

davidrunger:
Update:

image

Seems to have decreased worker memory usage relative to web somewhat.

PR #6613, opened 2025-03-30
[BlogController] Give 'sub!' replacement via block [BLOG-29]

davidrunger:
This avoids quite the little gotcha when a user's name includes a single quote:

https://rubyapi.org/3.4/o/string#class-String-label-Substitution+Methods :

> Note that within the string `replacement` [...] you may refer to some special match variables using these combinations: [...] `\'` corresponds to `$'`, which contains the string after the match.

We did not intend to have this behavior. The result was that a lot of the HTML would be copied into the `html` variable twice (e.g. it would have two closing `</html>` tags), and in general it messed up the HTML content.

Providing the replacement string instead as the return value of a block avoids this issue.

### DIFF
--- a/spec/features/blog_spec.rb
+++ b/spec/features/blog_spec.rb
@@ -52,82 +52,93 @@ RSpec.describe 'Blog' do
       context 'when the replying user is at least 1 minute old' do
         before { existing_user.update!(created_at: 1.minute.ago) }
 
-        it 'shows the blog article content and allows making comments, replying to comments, editing comments, and deleting comments' do
-          visit blog_url_path
+        context 'when the public_name of the user initially has a singe quote' do
+          let(:public_name) { "David O'Runger" }
 
-          expect(page).to have_text(h1_text)
-
-          click_sign_in_with_google
-          fill_in('Public display name (optional)', with: Faker::Name.name)
-          click_on('Submit')
-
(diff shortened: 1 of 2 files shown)


davidrunger:
# View ignoring whitespace-only changes

https://github.com/davidrunger/david_runger/pull/6613/files?w=1

spec/features/blog_spec.rb
```diff
@@ -52,82 +52,93 @@
...
-            find_all(:fillable_field, 'Write a comment...')[1].send_keys(additional_text)
-            click_on('Update')
+            click_on('Edit your name', match: :first)
+            new_name = Faker::Name.name
+            fill_in('Public display name (optional)', with: new_name)
+            click_on('Submit')
 
-            expect(page).to have_css('.comment', text: "#{reply_text}#{additional_text}")
-          end
+            expect(page).to have_text("Post as #{new_name}")
```
davidrunger, line 78: Most of the "diff" in this spec is just indenting things to accommodate the new `when the public_name of the user initially has a singe quote` context description. However, this part is new, because in addition to the hardcoded name with an apostrophe, I also want to continue using Faker to generate a name. That's how this bug was discovered, after all, so clearly there is some value in using the various names provided by Faker.

PR #6614, opened 2025-03-31
Add a universal logged-in header [DEV-17]

davidrunger:
Indicating to the user that they are logged in and giving them the ability to log out is a basic UX minimum. Users should also be able to access more information about their account and the ability to take actions on it (such as deleting their account).

For this reason, this change adds a header that will appear for logged in users on all pages. The header shows the user's email and also has a "My account" link and a "Sign out" button.

The header looks and works fine on mobile.

It was a bit of work to get it looking decent on all of our various apps (in both light mode and dark mode when relevant, such as on the blog), but I think that it does.

This change removes the logged-in menu info from the sidebar of the Groceries app, since that is now redundant with this header.

I am going with a header rather than a footer because I think that a header is more mobile friendly. On a phone, I think that the screen nearer to the bottom of the screen is more valuable to save for user interactions, since it is easier to reach with one's thumb. The higher parts of the screen are less valuable, since they are harder to reach. Therefore, I am putting the relatively low-interaction logged-in info etc into a header rather than a footer.

image


app/assets/stylesheets/styles.css
```diff
@@ -44,13 +44,9 @@
   }
 }
 
-html,
-body {
```
davidrunger, line 48: Having `body` here was causing an issue on the home page with the logged-in header having `position: sticky`. The header would scroll up with the hero. Removing `body` fixes that.

app/controllers/blog_controller.rb
```diff
@@ -100,8 +100,12 @@ def blog_html_with_additions(absolute_path)
 
     html.sub!('</body>') { "#{helpers.ts_tag('comments')}</body>" }
 
-    if params[:action] == 'show'
```
davidrunger, line 103: We now want to render Tailwind on the blog index page, as well, so that it can be used to style the logged-in header, which we will render there if a user is logged in.

app/javascript/groceries/Groceries.vue
```diff
@@ -106,6 +106,12 @@ function warnIfRequestPending(event: BeforeUnloadEvent) {
   touch-action: manipulation;
 }
 
+header {
+  color: var(--color-neutral-800);
+  background: var(--color-indigo-100);
+  border-bottom: 1px solid var(--color-neutral-300);
+}
```
davidrunger, line 113: Custom styling here for groceries because neither a dark them nor a light theme looks good.

app/javascript/home/Home.vue
```diff
@@ -77,6 +77,16 @@ function setScrollToFragmentTimeouts() {
...
 <style lang="scss">
+:root {
+  --user-header-height: 32px;
+  --main-bg-color: var(--color-neutral-950);
+  --main-text-color: var(--color-neutral-100);
+}
+
+.logged-out {
+  --user-header-height: 0px;
+}
```
davidrunger, line 88: This (and the related code change in `HomeHeader.vue` below) is a bit hacky, but it works to situate the logged-in header height and the navigation header as desired relative to each other, allowing for the fact that a user might not be logged in.

app/javascript/home/Home.vue
```diff
@@ -77,6 +77,16 @@ function setScrollToFragmentTimeouts() {
 </script>
 
 <style lang="scss">
+:root {
+  --user-header-height: calc(var(--spacing) * 8);
```
davidrunger, line 81: This is the same value as the `.h-8` on the logged-in header. Hopefully I will never change that without thinking to also update this. It would be great if I could afford a Percy snapshot of the home page for both the logged-out and logged-in cases, but I don't really think that I can / that that would be the best use of snapshots. It will probably be relatively rare for logged-in users to visit the home page, anyway.

app/javascript/logs/Logs.vue
```diff
@@ -1,17 +1,14 @@
 <template lang="pug">
-div
-  header.flex.justify-between.p-2(v-if="currentUser")
-    div {{ currentUser.email }}
```
davidrunger, line 4: The logs app showed the user's email, which would have been redundant with the new universal logged-in header, so we are removing it here.

app/views/layouts/_logged_in_header.html.haml
```diff
@@ -0,0 +1,5 @@
+%header.sticky.top-0.left-0.right-0.flex.justify-between.items-center.py-2.px-3.max-w-dvw.overflow-scroll.h-8.z-100.border-b-1.border-neutral-800.text-base{class: 'text-(--main-text-color) bg-(--main-bg-color,var(--background-color))'}
+  .flex-1.truncate{title: current_user.email}= current_user.email
+  %div.flex.gap-4
+    %div= link_to('My account', my_account_path, class: 'text-inherit!')
+    %div= button_to('Sign out', destroy_user_session_path, method: :delete)
```
davidrunger: Styling this logged-in header entirely with Tailwind allows us to render and style it on the blog without needing to pull in an additional stylesheet (beyond the Tailwind styles and other global styles that we were already pulling in for show pages since https://github.com/davidrunger/david_runger/pull/6599 and for the index page as of this PR).

lib/test/tasks/run_file_size_checks.rb
```diff
@@ -18,8 +18,8 @@ class Test::Tasks::RunFileSizeChecks < Pallets::Task
     'emoji_picker*.css' => (8..18),
     'emoji_picker*.js' => (322..332),
     'google_sign_in_button*.js' => (1..11),
-    'groceries*.css' => (60..70),
-    'groceries*.js' => (319..329),
+    'groceries*.css' => (48..58),
+    'groceries*.js' => (303..313),
```
davidrunger, line 22: The decreased file sizes here are a small but nice benefit of removing the groceries menu for account management.

app/views/layouts/_logged_in_header.html.haml
```diff
@@ -0,0 +1,5 @@
+%header.sticky.top-0.left-0.right-0.flex.justify-between.items-center.py-2.px-3.max-w-dvw.overflow-scroll.h-8.z-100.border-b-1.border-neutral-800.text-base{class: 'text-(--main-text-color,var(--text-color)) bg-(--main-bg-color,var(--background-color))'}
```
davidrunger, line 1: `--main-bg-color` is set in `Home.vue` and `Logs.vue`.

`--main-text-color` is set in `Home.vue`.

`--text-color` and `--background-color` are set by the blog.

app/javascript/groceries/components/StoreHeader.vue
```diff
@@ -1,5 +1,5 @@
 <template lang="pug">
-h1.h2.store-name.my-4
+h1.h2.my-1
```
davidrunger, line 2: By changing `my-4` to `my-1` , we are able to reclaim a lot of the vertical space otherwise lost by adding the new header.

PR #6620, opened 2025-04-01
Provide 'apple-mobile-web-app-status-bar-style' meta tag w/ 'black'

davidrunger:
This will supposedly make the status bar white text on a black background. I think that this will look a little better for the groceries app and especially the logs app.

Also, remove the `name: 'theme-color', content: '#ffffff'` meta tag, which seems like the opposite of our goal of moving toward a darker theme.

I don't want to bother with trying to test this locally, so I am just going to deploy to production and hope it works.

I based this change on a few articles I looked at, like this one: https://medium.com/appscope/changing-the-ios-status-bar-of-your-progressive-web-app-9fc8fbe8e6ab .

One interesting note is that, in iOS, for some reason the home page already has white text on a black background in the status bar (?), but other apps like logs and groceries do not.

I continue to dislike developing for iOS. They don't seem to have prominent documentation.


davidrunger:
This was for https://davidrunger.atlassian.net/browse/GROC-31 .

PR #6653, opened 2025-04-02
[comments] Allow editing public name inline [COM-10]

davidrunger:
This way, we won't redirect users to an ugly page that just has a field to edit their public name.

This change leverages the preexisting `useCancellableInput` composable.


app/controllers/my_account_controller.rb
```diff
@@ -18,26 +18,9 @@ def edit
...
-    authorize(current_user)
-    current_user.update!(my_account_params)
-    redirect_to(redirect_location)
-  end
-
-  private
-
-  def my_account_params
-    params.expect(user: %i[public_name])
-  end
```
davidrunger, line 42: These methods are added back in #6654 .

PR #6667, opened 2025-04-05
[specs/proposal] Delete spec that is redundant with a check-ins spec

davidrunger:
This change deletes a spec from `spec/features/proposal_spec.rb` that was mostly a redundant duplicate of a spec in `spec/features/check_ins_spec.rb`, to which I am adding a few lines that were present in the `spec/features/proposal_spec.rb` spec but not in the `spec/features/check_ins_spec.rb` spec. I am choosing to preserve the `spec/features/check_ins_spec.rb` spec because it has a section that adds an emotional need, which the `spec/features/proposal_spec.rb` spec did not.

### DIFF
--- a/spec/features/check_ins_spec.rb
+++ b/spec/features/check_ins_spec.rb
@@ -56,6 +56,9 @@ RSpec.describe 'Check-Ins app' do
             # rubocop:enable RungerStyle/ClickAmbiguously
             expect(page).to have_content('Marriage created.')
           end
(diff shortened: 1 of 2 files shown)


davidrunger:
Codecov seems to be having an issue right now, so I am going to merge this, even though the deletion of a spec would be an opportune time to have Codecov confirm that our coverage is still at 100%.

PR #6730, opened 2025-04-20
[dependabot] Use cron syntax to update dependencies every day

davidrunger:
https://github.com/dependabot/dependabot-core/issues/ 6339#issuecomment-2810799331


davidrunger:
https://davidrunger.atlassian.net/browse/DEV- 243

PR #6733, opened 2025-04-21
[emoji-picker] Redirect back to emoji picker after login [EMO-13]

davidrunger:
Also, provide a button to log in directly, rather than going through the `new_user_session_path`.

### DIFF
--- a/app/javascript/comments/Comments.vue
+++ b/app/javascript/comments/Comments.vue
@@ -5,7 +5,7 @@
   template(v-if="store.currentUser")
     CommentForm(@submit="store.addComment")
   template(v-else)
-    p #[GoogleLoginButton.google-login-form.dark-mode-supported(:origin="windowLocationWithHash('comments')")] to add a comment.
+    p #[GoogleLoginButton.dark-mode-supported(:origin="windowLocationWithHash('comments')")] to add a comment.
 
   .mt-8
     template(v-if="store.comments.length")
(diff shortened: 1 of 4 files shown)


app/javascript/comments/Comments.vue
```diff
@@ -5,7 +5,7 @@
   template(v-if="store.currentUser")
     CommentForm(@submit="store.addComment")
   template(v-else)
-    p #[GoogleLoginButton.google-login-form.dark-mode-supported(:origin="windowLocationWithHash('comments')")] to add a comment.
```
davidrunger, line 8: (Mostly unrelated change; I noticed when copying from here that this `.google-login-form` class is no longer needed, since this change: https://github.com/davidrunger/david_runger/pull/6599/files#diff-990c373ac35fdfc6f3e009894c7e0c44276f98f0bf472dcb50d9bbc0b9a31e28L53-L55 .)

app/javascript/emoji_picker/EmojiPicker.vue
```diff
@@ -66,6 +62,8 @@ const { highlightedSearchable, onArrowDown, onArrowUp, topRankedMatches } =
     },
   });
 
+const googleLoginOrigin = window.location.href;
```
davidrunger, line 65: (I'd have liked to simply inline this, but trying to do so was generating a TypeScript complaint.)

PR #6832, opened 2025-05-20
Update NPM packages

davidrunger:


package.json
```diff
@@ -79,26 +79,26 @@
...
+    "eslint-import-resolver-typescript": "^4.3.5",
     "eslint-plugin-import": "^2.31.0",
     "fishery": "^2.3.1",
     "globals": "^16.1.0",
     "jsdom": "^26.1.0",
     "postcss-html": "^1.8.0",
     "prettier": "^3.5.3",
-    "quicktype": "^23.1.4",
-    "sass": "^1.88.0",
+    "quicktype": "23.2.4",
```
davidrunger, line 89: Locking to this version to avoid this bug: https://github.com/glideapps/quicktype/issues/ 2753 .

tools/json_schemas_to_typescript.rb
```diff
@@ -61,7 +61,7 @@ def write_files(changed: nil, added: nil, removed: nil)
           if write_all || schema_path.in?(changed) || schema_path.in?(added)
             types_path = types_path(schema_path)
 
-            system(<<~SH.squish)
+            system(<<~SH.squish, exception: true)
```
davidrunger, line 64: This would have made the issue resulting from https://github.com/glideapps/quicktype/issues/ 2753 more clear / immediately visible.

package.json
davidrunger: I am pretty annoyed by quicktype. This is the second time in the past month that a pretty big/fundamental bug has prevented me from just bumping all dependencies. The other time was this bug: https://github.com/glideapps/quicktype/issues/ 2721 .

PR #6892, opened 2025-06-01
Alias 'vite' to 'rolldown-vite'

davidrunger:
This should make our JS bundling much faster!


davidrunger:
Still not working. :(

PR #7244, opened 2025-08-27
[home/projects/davidrunger.com] Update NGINX link

davidrunger:


davidrunger:
**Link to https://nginx.org/ did not return expected status**

We made a request to https://nginx.org/, which is linked from https://davidrunger.com/, and its response status was 307, but we expected it to be in [200]. 😕

PR #7520, opened 2025-10-14
[GitHub Actions] Prune pnpm store before cacheing

davidrunger:
This should make the cache size smaller and maybe faster. I'm not totally sure that this is a win, on net, though, since doing the pruning will take some time. But this feels better to me than having the cache grow indefinitely.


davidrunger:
It seems to be a net win.

Before (total 26 seconds):

image

After (total 12 seconds):

image

PR #7756, opened 2025-11-18
Update NPM packages

davidrunger:


davidrunger:
[Percy is down](https://status.percy.io/incidents/w0f2v4gtqs5y), so I merged without that check.

PR #7841, opened 2025-12-05
Source 'rails' from GitHub

davidrunger:


davidrunger:
I was trying this to fix https://github.com/davidrunger/david_runger/actions/runs/19972458342/job/57280544993 but it seems like a lot of changes will be required. I will lock connection_pool instead in a different PR.

PR #8109, opened 2026-01-20
[specs] Fix missing 'apples' by increasing timeout [GROC-37]

davidrunger:


davidrunger:
Relevant part of spec:

```rb
# Delete the item.
within('.grocery-item', text: unneeded_item.name) do
  find(:link_or_button, 'Delete item').click
  sleep(0.01)
  expect(page).not_to have_selector(:link_or_button, 'Delete item')
end

# Confirm that the deleted item is no longer listed.
expect(page).not_to have_css('.grocery-item', text: unneeded_item.name)

# Undo the deletion
click_on('Undo')

# Confirm that the item is listed again.
expect(page).to have_css('.grocery-item', text: unneeded_item.name)
```

We would delete apples, and then restore it. However, sometimes it wouldn't actually be restored.

The feature spec logs revealed that the server was not receiving a request to restore the apples item.

However, the `click_on('Undo')` wasn't complaining that the button wasn't there.

The toast in which that "Undo" button is rendered is animated out after a timeout. I think what was happening is that, sometimes, the click on "Undo" would be not so slow that the "Undo" button wasn't there at all, but somehow slow enough that the button no longer worked / wasn't correctly clicked (maybe because it was being animated out).

I tested this via something like `sleep(4.35)` before `click_on('Undo')`. This reproduced the flake in question maybe 25% of the time. Sometimes the "Undo" button would be missing altogether, producing a different error. Sometimes it would be clicked successfully. But sometimes it wouldn't complain that the button was not there, but we'd get the flaky failure that is the subject of the ticket which this PR aims to fix.

Increasing the timeout from 5 seconds (the default) to 60 seconds in tests should get rid of this flakiness.

PR #8111, opened 2026-01-20
[specs] Use separate Redis database numbers for Sidekiq [DEV-247]

davidrunger:
Thanks to DeepSeek for figuring this out: https://chat.deepseek.com/share/5uj3nzq696am69gf5r . (I wish I'd have figured this out on my own, and I am somewhat ashamed that I didn't. I think part of the reason I didn't think of it is that I knew I was using separate databases for the _application_ Redis connection. Unfortunately, it didn't really occur to me that we need to do the same thing for the _Sidekiq_ Redis connections, too.)

### DIFF
--- a/app/poros/redis_options.rb
+++ b/app/poros/redis_options.rb
@@ -1,8 +1,13 @@
 class RedisOptions
   prepend Memoization
 
-  def initialize(db: nil)
-    @db = db || default_db_number
+  def initialize(db: nil, sidekiq: false)
+    @db =
+      if rails_test?
+        test_db_number(sidekiq:)
+      else
+        db || 0
+      end
   end
 
   memoize \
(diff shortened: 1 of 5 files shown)


app/poros/redis_options.rb
```diff
@@ -21,19 +26,23 @@ def url_without_db
...
+  def test_db_number(sidekiq:)
+    # piggyback on the Postgres DB_SUFFIX ENV variable to choose a Redis DB number
+    base_db_number =
       case ENV.fetch('DB_SUFFIX', nil)
       when '_unit', nil then 4
       when '_api' then 5
       when '_html' then 6
       when '_feature_a' then 7
-      when '_feature_c' then 9
+      when '_feature_c' then 8
```
davidrunger, line 37: I had been leaving `8` available in case I add `_feature_b`, but I think we should just compact `_feature_c` down to 8, since we are starting to run up against the Redis DB limit of `15`.

PR #8120, opened 2026-01-21
[specs/home] Click resume link twice [HOME-23]

davidrunger:
With this change, in the home feature spec where it tests `external_link_click` event tracking, we will now click the resume link twice (rather than once, as before), because the event tracking as the page unloads is intrinsically flaky (not just in test, but also for real users, I think). Clicking the link twice makes the chance of flaky failures much smaller (and low enough that we might never see a flake again). We expect that two events will usually now be created. Rarely, one event might be created. Even more rarely, no events might be created, in which case this spec will fail. But, as mentioned, hopefully this will be so rare that we might never see this spec flake ever again.

### DIFF
--- a/spec/features/home_spec.rb
+++ b/spec/features/home_spec.rb
@@ -59,9 +59,15 @@ RSpec.describe 'Home page', :prerendering_disabled do
     # External link click tracking >>>
     event_count_before = Event.count
 
+    # Click the resume link twice, because the event tracking as the page
+    # unloads is intrinsically flaky (not just in test, but also for real).
+    # Clicking the link twice makes the chance of flaky failures much smaller
+    # (and low enough that we might never see a flake again).
+    click_on('View Resume (pdf)')
+    page.driver.go_back
     click_on('View Resume (pdf)')
 
-    wait_for { Event.count }.to eq(event_count_before + 1)
+    wait_for { Event.count }.to be_between(event_count_before + 1, event_count_before + 2)
 
     new_event = Event.reorder(:created_at).last!
 


davidrunger:
Unfortunately, this seems to introduce a different flakiness:

```
Failures:

  1) Home page / says "David Runger / Full stack web developer" and tracks scrolling and clicks on external links
     Got 1 failure and 1 other error:

     1.1) Failure/Error: click_on('View Resume (pdf)')

          Capybara::ElementNotFound:
            Unable to find link or button "View Resume (pdf)"
          # ./config/initializers/z.rb:577:in 'Method#call'
          # ./config/initializers/z.rb:577:in 'Capybara::DSL#click_on'
          # ./spec/features/home_spec.rb:68:in 'block (2 levels) in <top (required)>'
          # ./spec/spec_helper.rb:365:in 'block (5 levels) in <main>'
          # ./spec/spec_helper.rb:364:in 'UnboundMethod#bind_call'
          # ./spec/spec_helper.rb:364:in 'block (4 levels) in <main>'
          # ./spec/spec_helper.rb:362:in 'block (3 levels) in <main>'
          # ./spec/spec_helper.rb:361:in 'block (2 levels) in <main>'
          # ./spec/spec_helper.rb:344:in 'block (2 levels) in <main>'
          # ./spec/spec_helper.rb:190:in 'block (2 levels) in <main>'
          # ./spec/spec_helper.rb:181:in 'block (2 levels) in <main>'
          # -e:1:in 'Kernel#require'
          # -e:1:in '<main>'

     1.2) Failure/Error: expect(StoredWarnings.warnings).to be_empty

                                 Expected ["WARN: Screenshot could not be saved. `page.current_path` is empty.\n"]
            to return a truthy result for empty? or emptys?
          # ./spec/spec_helper.rb:183:in 'block (2 levels) in <main>'
          # -e:1:in 'Kernel#require'
          # -e:1:in '<main>'

Finished in 12.39 seconds (files took 1.44 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/features/home_spec.rb:2 # Home page / says "David Runger / Full stack web developer" and tracks scrolling and clicks on external links
```

I am just going to delete this flaky part of the spec. When the real behavior is flaky, the spec will inevitably flaky. Since I don't accept flaky specs, I'm just going to delete it.

PR #8383, opened 2026-03-13
Switch from 'lodash-es' to 'es-toolkit'

davidrunger:
https://github.com/toss/es-toolkit

`es-toolkit` is supposedly faster, produces smaller bundles, and has built-in types.

Recommended by Claude: https://claude.ai/chat/eb2b579e-f665-4f9d-8680-26882fbcd0e3


davidrunger:
This doesn't actually seem to decrease the bundle sizes (using `es-toolkit/compat`); it seems like overall a slight increase (some bundle sizes do decrease, but others increase):

```
RunFileSizeChecks : comments*.js size 345.04 is not in 346..356
RunFileSizeChecks : groceries*.js size 343.36 is not in 329..339
RunFileSizeChecks : logs*.js size 864.26 is not in 847..857
RunFileSizeChecks : workout*.js size 262.85 is not in 249..259
```

I guess I will merge it, anyway, though. I guess that, if we migrate off of `es-toolkit/compat`, maybe then the bundle sizes will decrease more? Also, supposedly `es-toolkit` is faster?

PR #8500, opened 2026-04-01
Use system Chromium for Percy

davidrunger:
This might make the Percy startup time faster, by never needing to download Chromium (?). Currently Percy startup (including, I think, Chromium download) is the bottleneck/blocker for starting system specs (since Vite moved from Rollup to the much faster Rolldown for JavaScript asset compilation). This should skip Chromium download, which will hopefully save a little time. I think that Chrome startup is still the slowest step, though, which, unfortunately, this doesn't help with. But I think it's worth seeing if we save a second or two, on average.

https://claude.ai/share/4de1324d-f882-4507-b3c5-3ec21f42c72e

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -62,15 +62,9 @@ jobs:
       - name: Set GIT_REV environment variable
         uses: ./.github/actions/git_rev
 
-      - name: Set PERCY_BROWSER_EXECUTABLE env var to system Chromium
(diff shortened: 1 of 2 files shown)


davidrunger:
I think that this is actually helping. On `main` after this merge, `StartPercy` took 10.1 seconds, whereas it took 14 - 17 seconds on the previous runs that I had looked at.

image

Also, we just got what I think might be an all-time best wall clock time of 57.83 seconds. It's not a ton faster than other recent runs, but it is faster by a second or two or maybe more, just as I had hoped that it might be.

image

PR #8802, opened 2026-06-05
Update NPM packages

davidrunger:


Gemfile.lock
```diff
@@ -1181,4 +1181,4 @@ RUBY VERSION
   ruby 4.0.5
 
 BUNDLED WITH
-  4.0.12
+  4.0.13
```
davidrunger, line 1184: Unrelated change. This should have been `4.0.13` for a while now, I think. I believe it was before, but then was changed back at some point.

PR #8818, opened 2026-06-09
[event] Require 'ip'

davidrunger:
All events should/will have an `ip`, so let's have our code and DB reflect and ensure that.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,30 +10,28 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.0].define(version: 2025_03_25_233233) do
-  create_schema "pghero"
-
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -10,55 +10,53 @@
...
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
   enable_extension "pg_stat_statements"
 
   create_table "action_mailbox_inbound_emails", force: :cascade do |t|
-    t.integer "status", default: 0, null: false
-    t.string "message_id", null: false
-    t.string "message_checksum", null: false
     t.datetime "created_at", null: false
+    t.string "message_checksum", null: false
```
davidrunger, line 20: It looks like the schema sorts the columns alphabetically, now, rather than by the order in which the columns were added. I think that this is probably a good change, but it generates some noise in this PR.

PR #8877, opened 2026-06-28
Migrate from eslint to oxlint

davidrunger:
Oxlint is insanely faster than eslint! It doesn't support everything that eslint supports (maybe some of which features were useful), but I am guessing it's good enough, especially since I'm not doing a whole lot of active development anymore. Also, I feel like I haven't been getting very much really useful feedback from eslint for a while, anyway.

Also, `eslint-plugin-import` has been slow to migrate from eslint 9 support to eslint 10, which is a bit of a concern and bother (especially since eslint 9 will be end-of-life in about 6 months, per https://github.com/import-js/eslint-plugin-import/issues/ 3227#issuecomment-4813681530). Oxlint was suggested as an alternative here: https://github.com/import-js/eslint-plugin-import/issues/ 3227#issuecomment-4817638228 .


.oxlintrc.json
```diff
@@ -0,0 +1,219 @@
+{
```
davidrunger, line 1: This file was generated via `npx @oxlint/migrate eslint.config.mjs` (and then formatted by Prettier). I don't think I tweaked anything myself (nor even really reviewed this file in any detail).

PR #8920, opened 2026-07-12
[Gemfile] Install 'benchmark' gem in all environments [DEV-256]

davidrunger:
Previously, we installed it only in development and test, but that resulted in a warning when installing only the production gems. I guess that some code that we are running in production (bundler code, I guess?) is requiring benchmark. This change stops that warning.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem 'activeadmin', '>= 4.0.0.beta15'
 gem 'addressable'
 gem 'alba'
 gem 'aws-sdk-s3'
+gem 'benchmark'
 gem 'blazer'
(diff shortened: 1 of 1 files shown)


davidrunger:
I think that this is because of a `Benchmark` reference in `z.rb`: https://github.com/davidrunger/dotfiles-personal/blob/37da0ac445cce002080b6857e202e3ef2b1556d7/initializers/z.rb#L412 . I think that's why we weren't seeing this warning in production deploys (e.g. here: https://github.com/davidrunger/david_runger/actions/runs/29196226482/job/86659769982#step:5:452). I sort of don't love adding `benchmark` to production just to avoid this issue locally, but whatever.

PR #8934, opened 2026-07-15
Update NPM packages

davidrunger:


package.json
```diff
@@ -77,16 +77,16 @@
...
+    "quicktype": "^23.3.25",
     "sass": "^1.101.0",
     "stylelint": "^17.14.0",
     "stylelint-config-standard": "^40.0.0",
     "stylelint-config-standard-scss": "^17.0.0",
     "stylelint-scss": "^7.2.0",
     "tsd": "^0.33.0",
-    "typescript": "^6.0.3",
-    "vite": "^8.1.3",
+    "typescript": "^7.0.2",
```
davidrunger, line 88: This is confusing/misleading. We appear to be bumping to 7.0.2, but, because of [the override added in this change](https://github.com/davidrunger/david_runger/pull/8934/changes#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794cR14-R15), we are actually effectively still on 6.0.3.

davidrunger: Fixed here: https://github.com/davidrunger/david_runger/pull/9139/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R88 .

PR #9150, opened 2026-08-15
[test] Reduce Pallets concurrency from 6 to 3 [DEV-273]

davidrunger:
The three feature RSpec workers and `RunHtmlControllerTests` normally overlap. Their CI timings have risen together, which suggests that CPU, memory, and PostgreSQL contention on the shared GitHub-hosted runner may be contributing to the recent slowdown.

Reduce `PALLETS_CONCURRENCY` from 6 to 3 so at most three of these heavyweight tasks run at once. This is an experiment: compare several CI runs by overall job wall-clock time as well as individual task durations before deciding whether to retain it.

The public `ubuntu-latest` runner is documented as a 4-CPU, 16-GB VM. The workflow also runs PostgreSQL, Redis, browser processes, Puma, and other test tasks, so its effective workload exceeds the number of available CPUs when all four RSpec tasks overlap.

https://docs.github.com/en/actions/reference/runners/github-hosted-runners

Wall clock run times on this branch:
1. 75.4 seconds (2 failures that might have added notable slowdown)
2. 83.3 seconds
3. 66.6 seconds
4. 95.6 seconds
5. 93.4 seconds
6. 94.9 seconds

- mean: 84.9 seconds
- median: 88.4 seconds

Note: this branch runs prettier (5.9 seconds on run 1), which `main` runs don't.

On passing runs on `main` since 2026-08-13 (28 runs):
- mean: 81.3 seconds
- median: 82.3 seconds

The first three runs looked promising, but the last 3 runs took the mean and median to worse than the baseline from `main`.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -112,7 +112,7 @@ jobs:
         run: bin/run-tests
         env:
           CI: true
-          PALLETS_CONCURRENCY: 6
+          PALLETS_CONCURRENCY: 3
           AWS_EC2_METADATA_DISABLED: true
           RAILS_ENV: test
           DISABLE_SPRING: 1
--- a/lib/test/requirements_resolver.rb
+++ b/lib/test/requirements_resolver.rb
@@ -251,8 +251,9 @@ class Test::RequirementsResolver
         !files_added_in?('spec/serializers')
     end,
     Test::Tasks::RunBrakeman => proc do
-      (!(haml_files_changed? || ruby_files_changed?) || running_locally?) &&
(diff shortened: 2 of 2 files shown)


davidrunger:
With Prettier suppressed, on this branch:
1. 80.6 seconds
2. 92.3 seconds
3. 92.3 seconds
4. 85.4 seconds

- mean: 87.65 seconds
- median: 88.85 seconds

davidrunger:
Doesn't seem to help, overall. Somewhat seems to hurt. Closing.

PR #9165, opened 2026-08-16
[test] Eager-load only unit specs in CI [DEV-279]

davidrunger:
Require both `CI` and `EAGER_LOAD=true` before eager loading the test environment instead of eager loading every Rails process in CI.

Set `EAGER_LOAD=true` only for `RunUnitTests`. The unit shard retains eager-loading validation, while API, HTML, feature, and other test subprocesses skip repeated eager loading during startup.


davidrunger:
Looks promising:

image

I'll merge.

PR #9188, opened 2026-08-17
Update JavaScript packages

davidrunger:


package.json
```diff
@@ -85,7 +85,7 @@
     "stylelint-config-standard-scss": "^17.0.0",
     "stylelint-scss": "^7.2.0",
     "tsd": "^0.33.0",
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
```
davidrunger, line 88: Unfortunately (because it's misleading, due to our override to use TypeScript 6), this is added back due to the new commands added here: https://github.com/davidrunger/dotfiles/pull/1373 .

PR #9207, opened 2026-08-20
[ci] Use a dedicated database for RunFeatureTestsB [DEV-294]

davidrunger:
Provision `david_runger_test_feature_b` in `CreateDbCopies` and run `RunFeatureTestsB` with `DB_SUFFIX=_feature_b`.

Give the new suffix its own `RedisOptions` database and remove the dependency on `RunApiControllerTests`, allowing FeatureTestsB to start as soon as its other prerequisites are complete.

Keep the Postgres suffix-to-Redis mapping in `TEST_DB_NUMBERS_BY_SUFFIX` and derive the Sidekiq offset from its size. This prevents Redis database collisions when the number of parallel spec databases changes, including the collision that would otherwise occur between `_unit` Sidekiq Redis and FeatureTestsC Redis.


davidrunger:
Here's the run that inspired this change:

image

PR #9221, opened 2026-08-21
[CI] Decrease pallets concurrency from 6 to 4

davidrunger:
As of #9215, I think that feature specs might be able to start running too early in CI, causing too much concurrent CPU load. I wonder if we might get faster CI runs by decreasing the concurrency enough to at least slow the starting of one of the feature spec batches until both unit specs and API controller specs have finished. This change implements that, at least as an experiment on this branch, and probably as an experiment on `main`, too.

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -122,7 +122,7 @@ jobs:
         run: bin/run-tests
         env:
           CI: true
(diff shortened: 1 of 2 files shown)


davidrunger:
Well, clock time on this branch run was 76.57, which is not so good, but also not so bad. Let's try it and see if maybe we get some faster runs on `main`. Merging.

PR #9252, opened 2026-08-23
[es-toolkit] Use native imports [DEV-298]

davidrunger:
Replace application imports from `es-toolkit/compat` with native `es-toolkit` APIs. Use native array methods for collection operations, adapt native signatures for `sortBy`, `pull`, and two-object `merge`, and preserve chart merges by merging into fresh targets.

The compatibility layer is no longer needed for the migrated call sites. Native `es-toolkit` includes deep merge support, so the chart code does not retain a compatibility exception. The CI-equivalent Vite build reduced the affected JavaScript bundles enough to update `Test::Tasks::RunFileSizeChecks::CONSTRAINTS`.

codex resume 01a02de2-f9fc-7fd1-990a-8705ef030bd8


davidrunger:
Overall, I'm not too impressed by `es-toolkit`. Bundle sizes are now probably about back to where they were with lodash. The API seems less convenient. Maybe it's faster?, but I sort of doubt it. I think this was a bad recommendation by Claude.

